PR #14, opened 2019-03-18
API Fixes to update to latest TLS API from aws-c-io

justinboswell:
*Issue #, if available:* https://github.com/awslabs/aws-iot-device-sdk-cpp-v2/issues/13

*Description of changes:* API Fixes to update to latest TLS API from aws-c-io


samples/jobs/describe_job_execution/main.cpp
```diff
@@ -186,7 +186,7 @@ int main(int argc, char *argv[])
     }
 
     auto connectionOptions = tlsCtx.NewConnectionOptions();
-    connectionOptions.server_name = endpoint.c_str();
+    connectionOptions.server_name = aws_string_new_from_c_str(aws_default_allocator(), endpoint.c_str());
```
bretambrose: This is way beyond the scope of this CR, but as a general whine, I dislike the way we don't do full cleanup .  Sure, the process ends, but given how people tend to use code samples (cut-and-paste-and-modify), anyone who uses this and embeds it somewhere other than main() is going to naturally have resource issues.  

I've seen this in some of the unit tests to, where we don't necessarily call clean_up on various os resource wrappers (or leave them locked, etc...).  Those will get the cut-and-paste treatment too.

PR #21, opened 2019-04-15
Gg discovery

JonathanHenson:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/discovery/include/aws/discovery/ConnectivityInfo.h
+++ b/discovery/include/aws/discovery/ConnectivityInfo.h
@@ -1,0 +1,45 @@
+#pragma once
+/* Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License").
+* You may not use this file except in compliance with the License.
+* A copy of the License is located at
+*
+*  http://aws.amazon.com/apache2.0
+*
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+
+* This file is generated
+*/
+#include <aws/crt/JsonObject.h>
+
+#include <aws/discovery/Exports.h>
+
+#include <aws/crt/JsonObject.h>
+#include <aws/crt/StlAllocator.h>
+
+namespace Aws
+{
+    namespace Discovery
+    {
+        class AWS_DISCOVERY_API ConnectivityInfo final
+        {
+          public:
+            ConnectivityInfo() = default;
+
+            ConnectivityInfo(const Crt::JsonView &doc);
+            ConnectivityInfo &operator=(const Crt::JsonView &doc);
+
+            Aws::Crt::Optional<Aws::Crt::String> ID;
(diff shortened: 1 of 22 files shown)


discovery/include/aws/discovery/GGGroup.h
```diff
@@ -0,0 +1,39 @@
...
+#include <aws/discovery/GGCore.h>
+
+namespace Aws
+{
+    namespace Discovery
+    {
+        class AWS_DISCOVERY_API GGGroup final
+        {
+          public:
+            GGGroup() = default;
```
justinboswell: They're GGGREAT!

I was going to suggest a namespace or the use of Greengrass, but then it turns out this is the nomenclature GG uses :\

discovery/source/DiscoveryClient.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+            Crt::StringStream ss;
+            ss << "greengrass-ats.iot." << clientConfig.region << ".amazonaws.com";
+
+            Crt::Io::TlsConnectionOptions tlsConnectionOptions = clientConfig.tlsContext->NewConnectionOptions();
+            uint16_t port = 443;
+
+            if (Crt::Io::TlsContextOptions::IsAlpnSupported())
+            {
+                tlsConnectionOptions.SetAlpnList("x-amzn-http-ca");
```
graebm: Does this need follow-up changes in aws-c-http to work?

discovery/source/DiscoveryClient.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+                        requestOptions.headerArray = &hostNameHeader;
+                        requestOptions.headerArrayLength = 1;
+
+                        requestOptions.onStreamOutgoingBody = nullptr;
+                        requestOptions.onIncomingHeaders =
+                            [](Crt::Http::HttpStream &, const Crt::Http::HttpHeader *, std::size_t) {};
+                        requestOptions.onIncomingHeadersBlockDone =
+                            [callbackContext](Crt::Http::HttpStream &stream, bool) {
+                                callbackContext->responseCode = stream.GetResponseStatusCode();
```
graebm: [optional] this callback could be null and you could just query the response-code in the onStreamComplete() callback

discovery/include/aws/discovery/ConnectivityInfo.h
```diff
@@ -0,0 +1,45 @@
...
+* A copy of the License is located at
+*
+*  http://aws.amazon.com/apache2.0
+*
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+
+* This file is generated
```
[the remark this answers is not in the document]
JonathanHenson: these aren't generated.

discovery/source/DiscoveryClient.cpp
JonathanHenson: yep, and done

discovery/source/DiscoveryClient.cpp
JonathanHenson: I wanted the ability to branch on it before writing to the stream. When we start generating this stuff I'll prolly revisit it.

PR #33, opened 2019-06-03
assert issues

graebm:
- Add missing `#include <assert.h>` to enum-handling cpp files.
- Change assert() -> AWS_ASSERT() in input-validating discovery code.
- Run formatter on discovery code.

### DIFF
--- a/jobs/source/RejectedErrorCode.cpp
+++ b/jobs/source/RejectedErrorCode.cpp
@@ -18,6 +18,8 @@
 #include <aws/crt/StlAllocator.h>
 #include <aws/crt/StringUtils.h>
 
+#include <assert.h>
+
 static const size_t INVALIDTOPIC_HASH = Aws::Crt::HashString("InvalidTopic");
 static const size_t INVALIDJSON_HASH = Aws::Crt::HashString("InvalidJson");
 static const size_t INVALIDREQUEST_HASH = Aws::Crt::HashString("InvalidRequest");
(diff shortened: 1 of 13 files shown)


jobs/source/RejectedErrorCode.cpp
```diff
@@ -18,6 +18,8 @@
 #include <aws/crt/StlAllocator.h>
 #include <aws/crt/StringUtils.h>
 
+#include <assert.h>
```
justinboswell: Why is this required? Why not just use AWS_ASSERT?

graebm: When transforming a string -> enum, if we don't recognize the string that the service returned, it does `assert(0); return static_cast<JobStatus>(-1);`

It's not something we want to be fatal in a release product. I asked Jonathan if he wanted the assert removed, or preserved and he wanted it preserved here.

Sorry for merging before I noticed this comment

Kevin-Tacheny: /home/kevin/deviceagentcpp-build/aws-iot-device-sdk-cpp-v2-prefix/src/aws-iot-device-sdk-cpp-v2/discovery/source/DiscoveryClient.cpp:32:47: error: ‘AWS_ASSERT’ was not declared in this scope

Still an issue I believe

graebm: This ought to work, the include chain is: `DiscoveryClient.cpp` -> `#include <aws/crt/Types.h>` -> `#include <aws/common/common.h>` -> `#include <aws/common/assert.inl>` ->`#define AWS_ASSERT`

Have you updated to [aws-crt-cpp v0.4.0](https://github.com/awslabs/aws-crt-cpp/releases/tag/v0.4.0)?

Kevin-Tacheny:
Yeah I think you are correct I had some mismatch and now on to the next issue.  I am still stuck on this linker error in our code: https://github.com/awslabs/aws-crt-cpp/issues/26

PR #36, opened 2019-06-04
continuous integration

graebm:
Set up continuous integration

### DIFF
--- a/codebuild/common-posix.sh
+++ b/codebuild/common-posix.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -e
+
+echo "Using CC=$CC CXX=$CXX"
+
+BUILD_PATH=/tmp/builds
+mkdir -p $BUILD_PATH
+INSTALL_PATH=$BUILD_PATH/install
+mkdir -p $INSTALL_PATH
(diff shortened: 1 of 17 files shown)


codebuild/common-posix.sh
```diff
@@ -0,0 +1,31 @@
...
+CMAKE_ARGS="-DCMAKE_PREFIX_PATH=$INSTALL_PATH -DCMAKE_INSTALL_PREFIX=$INSTALL_PATH $@"
+
+# If TRAVIS_OS_NAME is OSX, skip this step (will resolve to empty string on CodeBuild)
+if [ "$TRAVIS_OS_NAME" != "osx" ]; then
+    sudo apt-get install libssl-dev -y
+fi
+
+# build aws-crt-cpp
+pushd $BUILD_PATH
+git clone --branch v0.4.0 https://github.com/awslabs/aws-crt-cpp.git
```
justinboswell: Feels like we should add the aws-common-runtime folder to this project, and just do a deps build of cpp transitively, rather than try to keep all these scripts in sync

graebm: I entered a task to do this before we GA.
I just wanted to get some CI up today to ensure the builds weren't currently broken

PR #56, opened 2019-08-19
H1b

bretambrose:
Updates discovery client to use new HTTP Api

### DIFF
--- a/discovery/source/DiscoveryClient.cpp
+++ b/discovery/source/DiscoveryClient.cpp
@@ -14,6 +14,7 @@
 #include <aws/discovery/DiscoveryClient.h>
 
 #include <aws/crt/Types.h>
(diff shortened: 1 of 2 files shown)


discovery/source/DiscoveryClient.cpp
```diff
@@ -91,32 +91,33 @@ namespace Aws
                     std::shared_ptr<Crt::Http::HttpClientConnection> connection, int errorCode) {
                     if (!errorCode)
                     {
+                        auto request = Aws::Crt::MakeShared<Crt::Http::HttpRequest>(m_allocator, m_allocator);
```
graebm: there are a few places here that can fail with OOM. This MakeShared(), also the SetMethod(), SetPath(), and AddHeader() below

PR #96, opened 2020-02-28
Update issue templates

KaibaLopez:
Templates for issues


.github/ISSUE_TEMPLATE/---bug-report.md
```diff
@@ -0,0 +1,38 @@
...
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Logs/output**
+If applicable, add logs or error output.
+
+*REMEMBER TO SANITIZE YOUR PERSONAL INFO*
+
+```
+ApiHandle::InitializeLogging(...)
```
bretambrose: In their code it will show up as something like

```
apiHandle.InitializeLogging(...);
```

I don't know if there's a better way to communicate that the call isn't static but we don't know how they've created their ApiHandle instance.

.github/ISSUE_TEMPLATE/---bug-report.md
```diff
@@ -0,0 +1,38 @@
...
+---
+
+Confirm by changing [ ] to [x] below to ensure that it's a bug:
+- [ ] I've searched for [previous similar issues](https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/) and didn't find any solution
+  
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**Platform/OS you're using**
+MacOs/Linux/Windows
```
bretambrose: Might be worth explicitly calling out device/hardware properties as well.

.github/ISSUE_TEMPLATE/---bug-report.md
KaibaLopez: yea but i'm thinking this is more of a hint on how to do it rather than instructions, that should be on the readme or documentation.

justinboswell: Let's go with this, and see if it confuses people. If it does, we'll write it another way until it doesn't :)

Issue #88, opened 2020-01-23
questionable deallocation scheme in raw_pub_sub sample code


jsakwa:
The pointer to the stack memory on line 407  of the raw_pub_sub, that is later use to free the memory seems problematic. Couldn't this memory be overwritten by the next loop, resulting in multiple frees/memory leak etc.?

https://github.com/aws/aws-iot-device-sdk-cpp-v2/blob/master/samples/mqtt/raw_pub_sub/main.cpp

Would be nice if the the ByteBuf type were a pointer, or better yet a smart-pointer that auto-released itself, which would make such things impossible.

KaibaLopez:
Hi @jsakwa,
 I may be wrong here but that pointer is in essence a r to l value cast, which is not really doing dynamic allocation, so it wouldn't have a memory leak risk at least...

jsakwa:
@KaibaLopez thanks for the reply --  I think there may actually be two issues (assuming that onPublishComplete is invoked in the background by connection->Publish):
1. payload itself could be overwritten in the next loop before the related object is freed by onPublishComplete
2. the payload struct is on the stack, and will no longer be valid when the loop body "while(true)" is evaluated the second time, if the onPublishComplete passes the pointer, it will be stack memory that is no longer, valid. See stack overflow for an explaination of this concern: https://stackoverflow.com/questions/26556449/are-variables-inside-a-loop-while-or-for-disposed-after-the-loop-has-completed

KaibaLopez:
@jsakwa , yea you're right and I see your point.
So, the problem with using a smart pointer here is that ByteBuf is not really a C++ class but a kind of C struct, and because ByteBuf doesn't have a proper destructor the pointer wouldn't work too well there. 
The real solution would involve changing ByteBuf to be a proper class with a proper destructor, but that takes time and so while we get to that, I did a pull request with sort of workaround or a quicker solution to make this sample a bit better.
Thanks for pointing this out by the way!

jsakwa:
@KaibaLopez I'm not sure if the solution works, but haven't had time to review it thoroughly... something simpler that you could try as an alternative to adding a second mutex would be to capture payload by value:

```
        auto onPublishComplete = [payload](Mqtt::MqttConnection &, uint16_t packetId, int errorCode) mutable {
            aws_byte_buf_clean_up(&payload);
... 
```

Cheers,
Joshua

psiphi75:
[...]

PR #127, opened 2020-07-01
Samples fix

TingDaoK:
*Issue #, if available:*

Possible dead lock without the predicate, and the onSubAck will show the subscription success even if the errorCode raised.
Related to https://github.com/awslabs/aws-c-mqtt/issues/136
https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/88

*Description of changes:*

- add predicate for each conditionVariable.wait()
- changed the booleans in basic_sub_pub & raw_sub_pub to atomic
- Fixed the onSubAck callback to check the error code and QoS returned.

### DIFF
--- a/samples/greengrass/basic_discovery/main.cpp
+++ b/samples/greengrass/basic_discovery/main.cpp
@@ -190,8 +190,8 @@ int main(int argc, char *argv[])
     Aws::Iot::MqttClient mqttClient(bootstrap);
     std::shared_ptr<Mqtt::MqttConnection> connection(nullptr);
 
-    std::promise<bool> connectionFinishedPromise;
-    std::promise<bool> shutdownCompletedPromise;
+    std::promise<void> connectionFinishedPromise;
+    std::promise<void> shutdownCompletedPromise;
 
     discoveryClient->Discover(thingName, [&](DiscoverResponse *response, int error, int httpResponseCode) {
         if (!error && response->GGGroups)
@@ -254,7 +254,7 @@ int main(int argc, char *argv[])
                             if (!errorCode)
                             {
                                 fprintf(stdout, "Successfully subscribed to %s\n", topic.c_str());
-                                connectionFinishedPromise.set_value(true);
+                                connectionFinishedPromise.set_value();
                             }
                             else
                             {
@@ -271,7 +271,7 @@ int main(int argc, char *argv[])
                     }
                     else
                     {
-                        connectionFinishedPromise.set_value(true);
+                        connectionFinishedPromise.set_value();
                     }
                 }
                 else
@@ -297,7 +297,7 @@ int main(int argc, char *argv[])
 
             connection->OnDisconnect = [&](Mqtt::MqttConnection & /*connection*/) {
                 fprintf(stdout, "Connection disconnected. Shutting Down.....\n");
-                shutdownCompletedPromise.set_value(true);
+                shutdownCompletedPromise.set_value();
             };
 
             if (!connection->Connect(thingName.c_str(), false))
--- a/samples/greengrass/basic_discovery/main.cpp
+++ b/samples/greengrass/basic_discovery/main.cpp
@@ -190,10 +190,8 @@ int main(int argc, char *argv[])
     Aws::Iot::MqttClient mqttClient(bootstrap);
     std::shared_ptr<Mqtt::MqttConnection> connection(nullptr);
 
-    std::mutex semaphoreLock;
-    std::condition_variable semaphore;
-    std::atomic<bool> connectionFinished(false);
-    std::atomic<bool> shutdownCompleted(false);
+    std::promise<bool> connectionFinishedPromise;
+    std::promise<bool> shutdownCompletedPromise;
 
     discoveryClient->Discover(thingName, [&](DiscoverResponse *response, int error, int httpResponseCode) {
         if (!error && response->GGGroups)
@@ -256,8 +254,7 @@ int main(int argc, char *argv[])
                             if (!errorCode)
                             {
                                 fprintf(stdout, "Successfully subscribed to %s\n", topic.c_str());
-                                connectionFinished = true;
-                                semaphore.notify_one();
+                                connectionFinishedPromise.set_value(true);
                             }
                             else
                             {
@@ -274,8 +271,7 @@ int main(int argc, char *argv[])
                     }
                     else
                     {
-                        connectionFinished = true;
-                        semaphore.notify_one();
+                        connectionFinishedPromise.set_value(true);
                     }
                 }
                 else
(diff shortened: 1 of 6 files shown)


samples/jobs/describe_job_execution/main.cpp
```diff
@@ -245,52 +248,58 @@ int main(int argc, char *argv[])
...
-        conditionVariable.wait(uniqueLock);
+        conditionVariable.wait(uniqueLock, [&]() { return subAcked || jobExecutionRejectedCompleted; });
 
         DescribeJobExecutionRequest describeJobExecutionRequest;
         describeJobExecutionRequest.ThingName = thingName;
         describeJobExecutionRequest.JobId = jobId;
         describeJobExecutionRequest.IncludeJobDocument = true;
         Aws::Crt::UUID uuid;
         describeJobExecutionRequest.ClientToken = uuid.ToString();
+        std::atomic<bool> publishDescribeJobExeCompelted(false);
```
justinboswell, line 288: Completed

samples/jobs/describe_job_execution/main.cpp
```diff
@@ -226,17 +226,20 @@ int main(int argc, char *argv[])
 
         // This isn't absolutely neccessary but since we're doing a publish almost immediately afterwards,
         // to be cautious make sure the subscribe has finished before doing the publish.
+        std::atomic<bool> subAcked(false);
```
graebm, line 229: suggestion: C++ offers more synchronization primitives than aws-c-common offers.
We could use something simpler than this mutex/condition_variable/atomic combo, like:
https://en.cppreference.com/w/cpp/thread/promise
```
std::promise<void> subAcked;

subAcked.set_value(); // to notify as done

subAcked.get_future().wait(); // to wait on it

Comment (TingDaoK) [reply samples/jobs/describe_job_execution/main.cpp:229]: I am stuck by converting stuff like `conditionVariable.wait(uniqueLock, [&]() { return subAcked || jobExecutionAcceptedCompleted; });` to promise... It seems like `when_any(futures...)` is not supported by C++ 11.

Comment (TingDaoK) [reply samples/jobs/describe_job_execution/main.cpp:229]: A possible workaround I can think of is 
`conditionVariable.wait(uniqueLock, [&]() { return connectionSucceeded || connectionClosed; });` -> `if(connectionCompletedPromise.get_future().get() || connectionClosedPromise.get_future().get()){ ... }` ?

File: samples/greengrass/basic_discovery/main.cpp [hunk 3]
```diff
@@ -190,10 +190,8 @@ int main(int argc, char *argv[])
     Aws::Iot::MqttClient mqttClient(bootstrap);
     std::shared_ptr<Mqtt::MqttConnection> connection(nullptr);
 
-    std::mutex semaphoreLock;
-    std::condition_variable semaphore;
-    std::atomic<bool> connectionFinished(false);
-    std::atomic<bool> shutdownCompleted(false);
+    std::promise<bool> connectionFinishedPromise;
+    std::promise<bool> shutdownCompletedPromise;
```
graebm, line 194: trivial: these can be `std::promise<void>` since we don't do anything with the bool inside them

samples/mqtt/basic_pub_sub/main.cpp
```diff
@@ -309,20 +301,29 @@ int main(int argc, char *argv[])
...
+                    {
+                        fprintf(stdout, "Subscribe on topic %s on packetId %d Succeeded\n", topic.c_str(), packetId);
+                    }
+                }
+                subscribeFinishedPromise.set_value(true);
+            };
 
         connection->Subscribe(topic.c_str(), AWS_MQTT_QOS_AT_LEAST_ONCE, onPublish, onSubAck);
-        conditionVariable.wait(uniqueLock);
+        subscribeFinishedPromise.get_future().wait();
```
graebm, line 326: We don't do anything differently if the subscribe failed ... which is odd
probably should print/exit from the callback when we get the error_code, or exit after the wait() if the value returned is false

samples/mqtt/basic_pub_sub/main.cpp
```diff
@@ -360,15 +361,16 @@ int main(int argc, char *argv[])
         /*
          * Unsubscribe from the topic.
          */
+        std::promise<bool> unsubscribeFinishedPromise;
```
graebm, line 364: trivial: this doesn't use its `bool` could be `void`

samples/shadow/shadow_sync/main.cpp
```diff
@@ -280,18 +273,15 @@ int main(int argc, char *argv[])
...
-            subscribeDeltaAccepedCompleted = true;
-            conditionVariable.notify_one();
+            subscribeDeltaAcceptedCompletedPromise.set_value(true);
         };
 
         auto onDeltaUpdatedRejectedSubAck = [&](int ioErr) {
             if (ioErr != AWS_OP_SUCCESS)
             {
                 fprintf(stderr, "Error subscribing to shadow delta rejected: %s\n", ErrorDebugString(ioErr));
             }
```
graebm, line 283: should probably exit() after print, like in the other subacks

samples/shadow/shadow_sync/main.cpp
```diff
@@ -251,27 +248,23 @@ int main(int argc, char *argv[])
...
     if (connectionSucceeded)
     {
         Aws::Iotshadow::IotShadowClient shadowClient(connection);
 
-        std::atomic<bool> subscribeDeltaCompleted(false);
-        std::atomic<bool> subscribeDeltaAccepedCompleted(false);
-        std::atomic<bool> subscribeDeltaRejectedCompleted(false);
+        std::promise<bool> subscribeDeltaCompletedPromise;
+        std::promise<bool> subscribeDeltaAcceptedCompletedPromise;
+        std::promise<bool> subscribeDeltaRejectedCompletedPromise;
```
graebm, line 259: not using any of these bools, we just exit if the subscribe fails

samples/jobs/describe_job_execution/main.cpp
```diff
@@ -212,92 +206,83 @@ int main(int argc, char *argv[])
...
+        auto subAckHandler = [&](int) {
+            /* if error code returns it will be recorded by the other callback */
+            subAckedPromise.set_value();
         };
-
         auto subscriptionHandler = [&](DescribeJobExecutionResponse *response, int ioErr) {
             if (ioErr)
             {
                 fprintf(stderr, "Error %d occurred\n", ioErr);
-                conditionVariable.notify_one();
```
TingDaoK, line 240: [SubscribeToDescribeJobExecutionAccepted](https://github.com/aws/aws-iot-device-sdk-cpp-v2/blob/master/jobs/source/IotJobsClient.cpp#L150) subAckHandler will always be called. I think we can probably safely wait for that callback instead of either one.

Issue #38, opened 2019-06-05
Mqtt::ReturnCode Bug in sample


ravichandhraRean:
Hi @graebm / @JonathanHenson ,

the recent updates,
1. The Mqtt::ConnectionState enum has been removed from the latest version of the SDK
2. Intoruduced Mqtt::ReturnCode

If you look into this sample code both,
1. In onConnectionCompleted callback lambda it is treating if there is no errorCode means connection is successful and just printing the connection completed message with  returnCode
2. it is assuming that the returnCode is always gives succes but If we look into the mqtt.h file there is a possibility of getting returnCode with different errors(We tested it we got 4/5 sometimes)
3. Our further business logic depends on this initial mqtt connection status, So could you please look into this.
4. And also can you add onResumed , onInterrupted  methods support for this sample

```
/* Result of a connect request [MQTT-3.2.2.3]. */
enum aws_mqtt_connect_return_code {
    AWS_MQTT_CONNECT_ACCEPTED,
    AWS_MQTT_CONNECT_UNACCEPTABLE_PROTOCOL_VERSION,
    AWS_MQTT_CONNECT_IDENTIFIER_REJECTED,
    AWS_MQTT_CONNECT_SERVER_UNAVAILABLE,
    AWS_MQTT_CONNECT_BAD_USERNAME_OR_PASSWORD,
    AWS_MQTT_CONNECT_NOT_AUTHORIZED,
    /* reserved = 6 - 255 */
};
```

```
    /*
     * This will execute when an mqtt connect has completed or failed.
     */
    auto onConnectionCompleted = [&](Mqtt::MqttConnection &, int errorCode, Mqtt::ReturnCode returnCode, bool) {
        if (errorCode)
        {
            fprintf(stdout, "Connection failed with error %s\n", ErrorDebugString(errorCode));
            std::lock_guard<std::mutex> lockGuard(mutex);
            connectionSucceeded = false;
        }
        else
        {
            fprintf(stdout, "Connection completed with return code %d\n", returnCode);
            connectionSucceeded = true;
        }
        {
            std::lock_guard<std::mutex> lockGuard(mutex);
            connectionCompleted = true;
        }
        conditionVariable.notify_one();
    };
```

Thank you.

Regards,
Ravichandhra Palla.

ravichandhraRean:
Hi @JonathanHenson / @graebm ,

Any update on this.

Thank you.

PR #130, opened 2020-07-08
Update sample to consider the connection failed if a non-accepted mqtt return code is seen

bretambrose:
https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/38

* Updates pub sub samples to also treat non-accepted return codes as errors

### DIFF
--- a/samples/mqtt/basic_pub_sub/main.cpp
+++ b/samples/mqtt/basic_pub_sub/main.cpp
@@ -232,7 +232,7 @@ int main(int argc, char *argv[])
      * This will execute when an mqtt connect has completed or failed.
      */
     auto onConnectionCompleted = [&](Mqtt::MqttConnection &, int errorCode, Mqtt::ReturnCode returnCode, bool) {
-        if (errorCode)
+        if (errorCode || returnCode != AWS_MQTT_CONNECT_ACCEPTED)
         {
             fprintf(stdout, "Connection failed with error %s\n", ErrorDebugString(errorCode));
             std::lock_guard<std::mutex> lockGuard(mutex);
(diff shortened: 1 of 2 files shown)


justinboswell (approved):

samples/mqtt/basic_pub_sub/main.cpp
```diff
@@ -238,6 +238,12 @@ int main(int argc, char *argv[])
             std::lock_guard<std::mutex> lockGuard(mutex);
             connectionSucceeded = false;
         }
+        else if (returnCode != AWS_MQTT_CONNECT_ACCEPTED)
```
graebm, line 241: Even better idea:

in aws-c-mqtt, if the returnCode is non-zero then error_code should be non-zero.

We'd need to add a new aws-error like: `AWS_ERROR_MQTT_CONNECTION_REFUSED`

Here are the docs for return code: http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349257

They're all just specific reasons for a failed connection.

> If a server sends a CONNACK packet containing a non-zero return code it MUST then close the Network Connection [MQTT-3.2.2-5].

graebm (changes requested):

samples/mqtt/basic_pub_sub/main.cpp
bretambrose: That makes sense.  Maybe I can trick Dengke into doing it.

bretambrose:
NM, let's do Mike's suggestion and use Dengke's pr.

PR #155, opened 2020-09-09
Use the common ALPN protocol by default in raw_pub_sub.

kankri:
Use the same ALPN protocol in the sample as in [aws-crt-cpp/README.md](../../../awslabs/aws-crt-cpp/blob/master/README.md)
and [`Aws:Iiot::MqttClientConnectionConfigBuilder::Build()`](../../../awslabs/aws-crt-cpp/blob/master/source/iot/MqttClient.cpp#L264).

---


massi-ang:
The protocol names are specified by AWS IoT Core https://docs.aws.amazon.com/iot/latest/developerguide/protocols.html and both repositories should refer to it for clarity.

kankri:
I added a comment including a link to the protocol names documentation.

graebm (approved):

bretambrose:
Merged in https://github.com/aws/aws-iot-device-sdk-cpp-v2/pull/182

Issue #165, opened 2020-09-29
Windows builds fail if base directory is longer than 53 characters


EeroKurimo:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/) and didn't find any solution

**Known Issue**
- [x] I'm using ATS data type endpoint: the endpoint should look like `<prefix>-ats.iot.<region>.amazonaws.com`
  
**Describe the bug**
If you try build the release version of the sdk on Windows in a base directory longer than 53 characters, the build will fail because MSBuild doesn't support paths longer than 260 characters regardless of the `LongPathsEnabled` registry setting, giving the error:

```
FileTracker : error FTK1011: could not create the new file tracking log file: C:\data\aws-sdk-cpp-workspace_123456789_123456789_123\aws-iot-device-sdk-cpp-v2-build\aws-common-runtime\aws-crt-cpp\aws-common-runtime\aws-c-event-stream\bin\aws-c-event-stream-write-test-casedir\RelWithDebInfo\aws-c-ev.0D0EAE0E.tlog\link-cvtres.write.1.tlog. The system cannot find the path specified. [C:\data\aws-sdk-cpp-workspace_123456789_123456789_123\aws-iot-device-sdk-cpp-v2-build\aws-common-runtime\aws-crt-cpp\aws-common-runtime\aws-c-event-stream\bin\aws-c-event-stream-write-test-case.vcxproj]
```

This is slightly related to https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/157 which we worked around by doing the checkout elsewhere, but in this case with build directories it is often not that easy in e.g. CI build environments. If the base directory is even longer, say 60 characters, there's even more build errors, though all appear to be coming from `aws-c-event-stream` builds.

To avoid a lot of future hassle for people building this on Windows, possible suggestions to allow longer base directory names could be:

- Clone `aws-crt-cpp` directly under `aws-iot-device-sdk-cpp-v2` without the intermediate `aws-common-runtime` directory.
- Under `aws-crt-cpp` clone the dependencies under a shorter dir `aws-crt` instead of `aws-common-runtime`.
- Under `aws-c-event-stream` avoid repeating the repo name, e.g. use test dir `write-test-casedir` instead of `aws-c-event-stream-write-test-casedir`.

**Platform/OS/Device**
Windows 10

[...]

PR #171, opened 2020-10-05
use max-path as linting job

graebm:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,3 +16,16 @@ jobs:
       with:
         # List of extensions to check
         extensions: cpp,h
+
+  max-path:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


justinboswell (approved):

justinboswell:
Why 150 and not 260, or near there?

graebm:
> Why 150 and not 260, or near there?

Currently, the longest path is ~200, which only gives the customer ~60 characters of prefix, which they're saying is tough to work around in this issue https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/165. If we limit ourselves to 150 then customers have ~100 characters to work with, which seems good.

I'm working on getting it down to ~130, but need to touch a bunch of places to get it there...

graebm:
closing due to inactivity

PR #184, opened 2020-10-30
Add device defender

giordzaws:
*Description of changes:*

* Added DeviceDefender functionality as a top level feature directory with its own CMakeLists file.
* Added DeviceDefender tests to be generated within the devicedefender dir.
* Added aws-c-iot repo as a submodule.
* Removed manylinux1 docker image targets since they are being phased out and are missing some linux header files required by device defender.

A previous PR for adding this functionality to the CRT can be found [here](https://github.com/awslabs/aws-crt-cpp/pull/193) for context. The code was reviewed, but it was determined that the sdk v2 repo is the more suitable home for it. Most of the source code in this PR is the same, but some parts have been modified such as the aws_iotdevice library initialization. Reviewers should be sure to go over the source code again. 


devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -0,0 +1,173 @@
...
+namespace Aws
+{
+    namespace Crt
+    {
+
+    }
+
+    namespace Iotdevice
+    {
+        class AWS_IOTDEVICEDEFENDER_API DeviceApiHandle final
```
fufranci, line 23: Putting `DeviceApiHandle` in `Iotdevice` namespace will be a problem when adding secure tunneling code to this repo.

Secure tunneling code will have something similar to initialize the the aws-c-iot library (`aws_iotdevice_library_init`). 

One option is to move `DeviceApiHandle` to `Iotdevicedefenderv1` namespace. Secure tunneling will define its own `DeviceApiHandle` in ST's namespace.

Too bad there is no common library for both DD and ST.

giordzaws: Ok, I'll move the `DeviceApiHandle` class into the `Iotdevicedefenderv1` namespace.

The common library does seem like it would be the right solution.

devicedefender/source/DeviceDefender.cpp
```diff
@@ -0,0 +1,201 @@
...
+            toMove.m_allocator = nullptr;
+            toMove.m_status = ReportTaskStatus::Stopped;
+            toMove.m_taskConfig = {0};
+            toMove.m_owningTask = nullptr;
+            toMove.m_lastError = AWS_ERROR_UNKNOWN;
+        }
+
+        ReportTask &ReportTask::operator=(ReportTask &&toMove) noexcept
+        {
+            OnTaskCancelled = std::move(toMove.OnTaskCancelled);
```
fufranci, line 78: check against

```
if (this != &toMove)
```

And also free existing resource of `this` before moving `toMove` to `this`.

giordzaws: Thanks for that call out, I'll update this in a follow up commit.

devicedefender/source/DeviceDefender.cpp
```diff
@@ -0,0 +1,202 @@
...
+            this->cancellationUserdata = nullptr;
+        }
+
+        ReportTaskBuilder::ReportTaskBuilder(
+            Aws::Crt::Allocator *allocator,
+            std::shared_ptr<Crt::Mqtt::MqttConnection> mqttConnection,
+            Crt::Io::EventLoopGroup &eventLoopGroup,
+            const Crt::String &thingName)
+            : m_allocator(allocator), m_mqttConnection(mqttConnection), m_thingName(thingName),
+              m_eventLoopGroup(std::move(eventLoopGroup))
```
DavidOgunsAWS, line 146: I'm thinking this shouldn't be a move. Why not just assign by ref?

devicedefender/source/DeviceDefender.cpp
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        ReportTask::ReportTask(
+            Aws::Crt::Allocator *allocator,
+            std::shared_ptr<Crt::Mqtt::MqttConnection> mqttConnection,
+            const Crt::String &thingName,
+            Crt::Io::EventLoopGroup &eventLoopGroup,
+            ReportFormat reportFormat,
+            uint64_t taskPeriodSeconds,
+            uint64_t networkConnectionSamplePeriodSeconds,
```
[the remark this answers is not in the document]
giordzaws: That makes sense, I'll update these to uint32_t

devicedefender/source/DeviceDefender.cpp
giordzaws: Good point, I'll update this and avoid using the move operation.

DavidOgunsAWS (approved):

PR #185, opened 2020-11-02
Add Secure Tunneling

fufranci:
*Description of changes:*
* Add secure tunneling as top level feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ endif()
 
 if (BUILD_DEPS)
     list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/crt/aws-crt-cpp/crt/aws-c-common/cmake")
+    list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/crt/aws-c-iot/cmake")
 
     include(AwsFindPackage)
     set(IN_SOURCE_BUILD ON)
@@ -61,6 +62,7 @@ add_subdirectory(discovery)
 add_subdirectory(identity)
 add_subdirectory(iotdevicecommon)
 add_subdirectory(devicedefender)
+add_subdirectory(secure_tunneling)
 
 if (BUILD_SAMPLES)
     message(WARNING "BUILD_SAMPLES has been deprecated. Please build each sample separately.")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,6 @@ endif()
 
 if (BUILD_DEPS)
     list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/crt/aws-crt-cpp/crt/aws-c-common/cmake")
-    list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/crt/aws-c-iot/cmake")
 
     include(AwsFindPackage)
     set(IN_SOURCE_BUILD ON)
--- a/secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
+++ b/secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
@@ -7,6 +7,7 @@
 #include <aws/crt/io/Bootstrap.h>
 #include <aws/crt/io/SocketOptions.h>
 #include <aws/iotdevice/secure_tunneling.h>
+#include <aws/iotsecuretunneling/Exports.h>
 
 int s_SecureTunnelingCallbackTest(Aws::Crt::Allocator *allocator, void *ctx);
 
@@ -24,7 +25,7 @@ namespace Aws
             using OnSecureTunnelingStreamReset = std::function<void(void)>;
             using OnSecureTunnelingSessionReset = std::function<void(void)>;
 
-            class SecureTunnel
+            class AWS_IOTSECURETUNNELING_API SecureTunnel
             {
               public:
                 SecureTunnel(
--- a/secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
+++ b/secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
@@ -27,12 +27,12 @@ namespace Aws
         };
 
         // Client callback type definitions
-        using OnSecureTunnelingConnectionComplete = std::function<void(void)>;
-        using OnSecureTunnelingSendDataComplete = std::function<void(int errorCode)>;
-        using OnSecureTunnelingDataReceive = std::function<void(const Crt::ByteBuf &data)>;
-        using OnSecureTunnelingStreamStart = std::function<void()>;
-        using OnSecureTunnelingStreamReset = std::function<void(void)>;
-        using OnSecureTunnelingSessionReset = std::function<void(void)>;
+        using OnConnectionComplete = std::function<void(void)>;
+        using OnSendDataComplete = std::function<void(int errorCode)>;
+        using OnDataReceive = std::function<void(const Crt::ByteBuf &data)>;
+        using OnStreamStart = std::function<void()>;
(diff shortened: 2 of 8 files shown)


secure_tunneling/source/SecureTunnel.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+            OnSecureTunnelingConnectionComplete onConnectionComplete,
+            OnSecureTunnelingSendDataComplete onSendDataComplete,
+            OnSecureTunnelingDataReceive onDataReceive,
+            OnSecureTunnelingStreamStart onStreamStart,
+            OnSecureTunnelingStreamReset onStreamReset,
+            OnSecureTunnelingSessionReset onSessionReset)
+        {
+            // Client callbacks
+            m_OnConnectionComplete = onConnectionComplete;
```
justinboswell, line 33: If you used `std::bind` here, you wouldn't need to unwrap `this` from the user data in the static functions, you could just call them.

secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
```diff
@@ -0,0 +1,100 @@
...
+            DeviceApiHandle(Crt::Allocator *allocator) noexcept;
+            ~DeviceApiHandle();
+            DeviceApiHandle(const DeviceApiHandle &) = delete;
+            DeviceApiHandle(DeviceApiHandle &&) = delete;
+            DeviceApiHandle &operator=(const DeviceApiHandle &) = delete;
+            DeviceApiHandle &operator=(DeviceApiHandle &&) = delete;
+        };
+
+        // Client callback type definitions
+        using OnSecureTunnelingConnectionComplete = std::function<void(void)>;
```
justinboswell, line 30: I don't think these names need to repeat the `SecureTunneling` part, they're already namedspaced.

secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
```diff
@@ -0,0 +1,100 @@
...
+#include <aws/iotdevice/secure_tunneling.h>
+#include <aws/iotsecuretunneling/Exports.h>
+
+int s_SecureTunnelingCallbackTest(Aws::Crt::Allocator *allocator, void *ctx);
+
+namespace Aws
+{
+    namespace Iotsecuretunneling
+    {
+        class AWS_IOTSECURETUNNELING_API DeviceApiHandle final
```
justinboswell, line 18: This doesn't beloing in Iotsecuretunneling, it belongs in a top level Iot namespace most likely. This will apply to DD as well.

secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
```diff
@@ -0,0 +1,100 @@
...
+            int SendData(const Crt::ByteCursor &data);
+
+            int SendStreamStart();
+
+            int SendStreamReset();
+
+            aws_secure_tunnel *GetUnderlyingHandle();
+
+          private:
+            void free();
```
justinboswell, line 77: This doesn't need to be a separate function, you can just call `this->~SecureTunnel()` in the move ctor.

secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
fufranci: Done. Thanks for pointing this out.

secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
fufranci: Done. I didn't know I can call the destructor directly. Thanks!

secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
fufranci: DD and ST are built as separate libraries. What happen when the customer uses both libraries. If I put this in the Iot namespace, I would need to name `DeviceApiHandle` to `SecureTunnelingDeviceApiHandle`, right?

giordzaws: Do we need a top level iotdevice folder that can declare the DeviceAPIHandle so that DD and ST can share that?

secure_tunneling/source/SecureTunnel.cpp
fufranci: I tried this:
```
m_OnConnectionComplete = std::bind(onConnectionComplete, this);
```
But my C++ is not strong enough to make this to work? In particular, what should be the type of `m_OnConnectionComplete` when I declare it in the header file? 

I also wanted to point out that `m_OnConnectionComplete` is a function pointer to a client supplied static function. When I call `m_OnConnectionComplete`
```
            auto *secureTunnel = static_cast<SecureTunnel *>(user_data);
            secureTunnel->m_OnConnectionComplete();
```
It looks like I am calling a member function of `SecureTunnel`, but it is not. This is calling a function pointer (to a client supplied function) which is a data member of `SecureTunnel`.

justinboswell: Oh, good call, nevermind then. You're right.

secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
justinboswell: > Do we need a top level iotdevice folder that can declare the DeviceAPIHandle so that DD and ST can share that?

That seems like the right move. This is where non-generated code would go that is common to IoT

justinboswell (approved):

secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
fufranci: The work is done in this PR:
https://github.com/aws/aws-iot-device-sdk-cpp-v2/pull/184

CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ endif()
 
 if (BUILD_DEPS)
     list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/crt/aws-crt-cpp/crt/aws-c-common/cmake")
+    list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/crt/aws-c-iot/cmake")
```
fufranci, line 45: @giordzaws Do you know if I need this line? I don't remember I added this line. Maybe it is a result of a rebase.

giordzaws: I don't think that line is required.

secure_tunneling/source/SecureTunnel.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+        int SecureTunnel::SendStreamStart() { return aws_secure_tunnel_stream_start(m_secure_tunnel); }
+
+        int SecureTunnel::SendStreamReset() { return aws_secure_tunnel_stream_reset(m_secure_tunnel); }
+
+        aws_secure_tunnel *SecureTunnel::GetUnderlyingHandle() { return m_secure_tunnel; }
+
+        void SecureTunnel::s_OnConnectionComplete(void *user_data)
+        {
+            auto *secureTunnel = static_cast<SecureTunnel *>(user_data);
```
DavidOgunsAWS, line 105: reinterpret_cast is intended for pointer conversions. You're not intending for a possible `SecureTunnel* operator()(void *ptr)` to handle it.

DavidOgunsAWS (approved):

secure_tunneling/source/SecureTunnel.cpp
fufranci: I will update these with a followup PR. Just want to merge this code for now to unblock UA development.

PR #195, opened 2020-11-17
Make copy if need to pass address to C layer

fufranci:
For memory that will be passed into the C layer, make a copy and store within the C++ secure tunnel object.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/secure_tunneling/CMakeLists.txt
+++ b/secure_tunneling/CMakeLists.txt
@@ -94,7 +94,6 @@ if (NOT IS_SUBDIRECTORY_INCLUDE)
     aws_use_package(aws-crt-cpp)
 endif()
 
-target_link_libraries(IotSecureTunneling-cpp ${DEP_AWS_LIBS})
 aws_use_package(aws-c-iot)
 
 if (BUILD_TESTING)
--- a/secure_tunneling/source/SecureTunnel.cpp
+++ b/secure_tunneling/source/SecureTunnel.cpp
@@ -12,11 +12,11 @@ namespace Aws
         SecureTunnel::SecureTunnel(
             Crt::Allocator *allocator,
             Aws::Crt::Io::ClientBootstrap *clientBootstrap,
-            Aws::Crt::Io::SocketOptions *socketOptions,
+            const Aws::Crt::Io::SocketOptions &socketOptions,
 
-            std::string accessToken,
+            const std::string &accessToken,
             aws_secure_tunneling_local_proxy_mode localProxyMode,
-            std::string endpointHost,
+            const std::string &endpointHost,
 
             OnConnectionComplete onConnectionComplete,
             OnSendDataComplete onSendDataComplete,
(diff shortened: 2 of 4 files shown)


secure_tunneling/CMakeLists.txt
```diff
@@ -94,7 +94,6 @@ if (NOT IS_SUBDIRECTORY_INCLUDE)
     aws_use_package(aws-crt-cpp)
 endif()
 
-target_link_libraries(IotSecureTunneling-cpp ${DEP_AWS_LIBS})
```
fufranci, line 97: See line 103. This line seems to be a duplicate.

secure_tunneling/source/SecureTunnel.cpp
```diff
@@ -33,17 +33,21 @@ namespace Aws
...
 
             config.allocator = allocator;
             config.bootstrap = clientBootstrap ? clientBootstrap->GetUnderlyingHandle() : nullptr;
-            config.socket_options = socketOptions ? &socketOptions->GetImpl() : nullptr;
+            config.socket_options = &m_socketOptions.GetImpl();
 
-            config.access_token = aws_byte_cursor_from_c_str(accessToken.c_str());
+            config.access_token = aws_byte_cursor_from_c_str(m_accessToken.c_str());
             config.local_proxy_mode = localProxyMode;
-            config.endpoint_host = aws_byte_cursor_from_c_str(endpointHost.c_str());
```
fufranci, line 46: The old code stores pointers to parameters which will go out-of-scope when this function returns.

secure_tunneling/source/SecureTunnel.cpp
```diff
@@ -102,37 +106,37 @@ namespace Aws
 
         void SecureTunnel::s_OnConnectionComplete(void *user_data)
         {
-            auto *secureTunnel = static_cast<SecureTunnel *>(user_data);
+            auto *secureTunnel = reinterpret_cast<SecureTunnel *>(user_data);
```
justinboswell, line 109: Why are these `reinterpret_cast` now? You can `static_cast` just fine from `void*` -> known type

secure_tunneling/tests/SecureTunnelTest.cpp
```diff
@@ -78,7 +79,7 @@ static int before(struct aws_allocator *allocator, void *ctx)
     testContext->secureTunnel = new Aws::Iotsecuretunneling::SecureTunnel(
         allocator,
         nullptr,
-        nullptr,
+        s_socketOptions,
```
justinboswell, line 82: If you're not going to configure this, you could just pass in `Aws::Crt::Iot::SocketOptions()`

justinboswell (approved):
Fix-n-ship

secure_tunneling/source/SecureTunnel.cpp
fufranci: It was suggested by @DavidOgunsAWS a while ago but I never submitted a PR for it. I honestly haven't read about when I should use which.

secure_tunneling/tests/SecureTunnelTest.cpp
fufranci: Ok. Will do.

secure_tunneling/source/SecureTunnel.cpp
fufranci: Undo the change and use `static_cast`.

PR #216, opened 2021-01-05
add new code example folder for basic pub sub with Kerberos straight strategy

ajainaus:
*Issue #, if available:*

*Description of changes:*


bretambrose:
We're in the process of reevaluating our approach to samples and potentially adding a number of targeted ones, but I think when the only variance is the authentication strategy, it still makes sense to have it be a CLI control off of a more generic sample.

PR #234, opened 2021-02-11
First attempt at dealing with sibling dependencies

bretambrose:
I initially attempted to add the local path to the dependency's cmake file to CMAKE_MODULE_PATH, but that didn't help.  If that's the right direction and I was doing it wrong, please advise.

### DIFF
--- a/devicedefender/CMakeLists.txt
+++ b/devicedefender/CMakeLists.txt
@@ -96,11 +96,7 @@ endif()
 
 aws_use_package(aws-c-iot)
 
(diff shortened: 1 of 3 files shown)


rpcme:
@bretambrose please see this recipe where CMAKE_MODULE_PATH does work.

https://github.com/aws/meta-aws/blob/gatesgarth/recipes-sdk/aws-iot-device-sdk-cpp-v2/aws-iot-device-sdk-cpp-v2_1.10.5.bb

PR #253, opened 2021-04-08
Modified the behavior of OnSubscribeComplete in secure tunneling samp…

aws-tec:
https://github.com/aws-tec/aws-iot-device-sdk-cpp-v2/blob/9274b93878ff33364710775ef4bb06590b26c02e/samples/secure_tunneling/main.cpp#L253

*Description of changes:*
In secure tunneling sample, it will always disconnect after OnSubscribeComplete callback. Hence it will never receive the message from AWS IoT core to notify tunneling creation. In this PR, add subscription successfully information instead of calling MQTT disconnect.


bretambrose:
Thanks for the PR.  We've refactored the samples and this particular one has been moved and a similar change to not disconnect was applied a while back.

PR #261, opened 2021-04-27
Eventstream RPC Client Connection

yourslab:
Some notes with regards to this implementation:
- The source code is located in a top-level `eventstreamrpc` directory. My current plan is to eventually have a separate directory for `greengrassipc`. However, it might also be worth considering having everything in a single directory.
- The connect amender returns a reference to a `MessageAmendment` object, and `std::move` is used with the original object erased from the headers list of the reference.
- `Crt::ScopedResource` is used to create a unique pointer with the AWS allocator for each created `EventstreamRpcConnection` object. This reduces the overhead when compared to using a shared pointer but also necessitates that the user rely on the invocation of the `OnError` and `OnDisconnect` callbacks to know when the pointer is no longer safe to use.
- `EventStreamHeader::operator==` does a lowercase comparison on the header name but also assumes that it is in ASCII. I wasn't quite sure on this one. It might be worth using `strcasecmp` for POSIX and `stricmp` for Windows.

Please feel free to leave any feedback. Thank you.


eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -0,0 +1,278 @@
...
+            void SetValue(bool value);
+            void SetValue(int8_t value);
+            void SetValue(int16_t value);
+            void SetValue(int32_t value);
+            void SetValue(int64_t value);
+            void SetValue(Crt::DateTime &value);
+            void SetValue(Crt::ByteBuf value);
+            void SetValue(Crt::UUID value);
+
+            const struct aws_event_stream_header_value_pair *GetUnderlyingHandle() const;
```
DavidOgunsAWS, line 122: The useful function that you want is to go from a C++ `EventStreamHeader` to the C `aws_event_stream_header_value_pair`. To do this, write an explicit conversion operator member function:

`explicit operator aws_event_stream_header_value_pair() const`

I'm fairly certain C++11 rules suggest return value optimization will occur on C structs so write it naively. That is return the stack allocated aws_event_stream_header_value_pair created inside the conversion by value. Don't try to move it or anything in the return statement.

eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -0,0 +1,278 @@
...
+
+            bool operator==(const EventstreamHeader &other) const noexcept;
+
+          private:
+            Crt::Allocator *m_allocator;
+            Crt::ByteBuf m_valueByteBuf;
+            struct aws_event_stream_header_value_pair m_underlyingHandle;
```
DavidOgunsAWS, line 129: Remove this in favor of managing the state of headers with C++ only types. Only when the C object is needed should the explicit conversion operator be called

eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -0,0 +1,278 @@
...
+            Crt::String HostName;
+            uint16_t Port;
+            OnConnect OnConnectCallback;
+            OnDisconnect OnDisconnectCallback;
+            OnError OnErrorCallback;
+            OnPing OnPingCallback;
+            ConnectMessageAmender ConnectMessageAmenderCallback;
+        };
+
+        class AWS_EVENTSTREAMRPC_API EventstreamRpcConnection
```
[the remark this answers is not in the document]
yourslab: I completely agree. I'm not entirely sure why HTTP took this approach either.

PR #266, opened 2021-05-06
guard include directory statements with condition on whether or not d…

rpcme:
…ependencies are being built

*Issue #, if available:*

Not available.

*Description of changes:*

Make guards for dependent cmake scripts for subcomponents so when you run a build with all dependencies already installed you can just use those - like a Yocto based build.

### DIFF
--- a/devicedefender/CMakeLists.txt
+++ b/devicedefender/CMakeLists.txt
@@ -90,11 +90,12 @@ target_include_directories(IotDeviceDefender-cpp PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
         $<INSTALL_INTERFACE:include>)
 
-if (NOT IS_SUBDIRECTORY_INCLUDE)
-    aws_use_package(aws-crt-cpp)
+if (BUILD_DEPS)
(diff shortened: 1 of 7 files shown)


jmklix:
can this be applied more generally for other statements?

justinboswell (commented):
Can you just add a comment explaining why this is here for the next unsuspecting person to end up in here? Otherwise, LGTM.

rpcme:
Hello @justinboswell I couldn't see a common mechanism because there is a mix a things that required guards across all those cmakefiles.  OTOH I am not in CMake every day so it could be lost on me.

rpcme:
> can this be applied more generally for other statements?

I couldn't see a good pattern other than creating new flags which only makes it more complex. I think reusing the flag and having a per-cmakefile change creates the least friction for the person running the cmakefile

sdavtaker (approved):
I think commit messages can be improved, but the rebase in the middle is making it harder to address it now.

sdavtaker (approved):

rpcme:
@justinboswell can this be pulled in yet?

justinboswell (approved):

PR #270, opened 2021-05-14
Greengrass IPC

yourslab:
- Fixes the definition of the flush callback along with futures returning the expected result of the operation
- Introduces a few mechanisms such as `ProtectedPromise` and mutexes to handle synchronization across multiple threads
- Avoids usage of pointers wherever possible in place of references

### DIFF
--- a/eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
+++ b/eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
@@ -258,7 +258,7 @@ namespace Aws
                 const Crt::Optional<Crt::ByteBuf> &payload,
                 OnMessageFlushCallback onMessageFlushCallback) noexcept;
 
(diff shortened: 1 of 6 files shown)


eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -190,44 +190,77 @@ namespace Aws
...
+            EVENT_STREAM_RPC_CRT_ERROR
+        };
+
+        struct EventStreamRpcStatus
+        {
+            EventStreamRpcError baseStatus;
+            int crtError;
+        };
+
+        template <class T> class ProtectedPromise
```
yourslab, line 214: Note that this type will never be public facing and is just a helper class. Ultimately, we still return an `std::future`.

yourslab:
Closing for #273

PR #273, opened 2021-05-14
GG IPC

yourslab:
- Fixes the definition of the flush callback along with futures returning the expected result of the operation
- Introduces a few mechanisms such as `ProtectedPromise` and mutexes to handle synchronization across multiple threads
- Avoids usage of pointers wherever possible in place of references
### TODO
- When the connection is no longer active or valid, trying to open a new stream or active an operation should return an error
- Code gen will be updated to have setters and getters for each member rather than having multiple constructors for each operation payload

### DIFF
--- a/eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
+++ b/eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
@@ -7,6 +7,7 @@
 #include <aws/eventstreamrpc/Exports.h>
 
 #include <aws/crt/DateTime.h>
+#include <aws/crt/JsonObject.h>
 #include <aws/crt/StlAllocator.h>
 #include <aws/crt/Types.h>
 #include <aws/crt/UUID.h>
@@ -14,7 +15,10 @@
 #include <aws/crt/io/SocketOptions.h>
 #include <aws/crt/io/TlsOptions.h>
 
+#include <aws/crt/io/HostResolver.h>
+
 #include <aws/event-stream/event_stream_rpc_client.h>
+#include <aws/io/host_resolver.h>
 
 #include <atomic>
 #include <functional>
@@ -34,9 +38,10 @@ namespace Aws
     {
         class EventStreamHeader;
         class EventStreamRpcClient;
+        class MessageAmendment;
         class ClientConnection;
         class ClientContinuation;
-        class MessageAmendment;
+        class ClientOperation;
 
         using HeaderValueType = aws_event_stream_header_value_type;
         using MessageType = aws_event_stream_rpc_message_type;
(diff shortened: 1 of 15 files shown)


eventstreamrpc/source/EventStreamClient.cpp
```diff
@@ -0,0 +1,1226 @@
...
+            m_closedPromise = std::move(rhs.m_closedPromise);
+            m_onConnectRequestCallback = rhs.m_onConnectRequestCallback;
+
+            return *this;
+        }
+
+        ClientConnection::ClientConnection(
+            ConnectionLifecycleHandler &connectionLifecycleHandler,
+            Crt::Allocator *allocator) noexcept
+            : m_allocator(allocator), m_lifecycleHandler(connectionLifecycleHandler)
```
DavidOgunsAWS, line 213: Probably need to initialize a few more states. At the very least, m_clientState needs to reflect DISCONNECTED and m_underlyingConnection should be nullptr.

eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -0,0 +1,605 @@
...
+            EVENT_STREAM_RPC_CONNECTION_CLOSED_BEFORE_CONNACK,
+            EVENT_STREAM_RPC_UNKNOWN_PROTOCOL_MESSAGE,
+            EVENT_STREAM_RPC_UNMAPPED_DATA,
+            EVENT_STREAM_RPC_UNSUPPORTED_CONTENT_TYPE,
+            EVENT_STREAM_RPC_STREAM_CLOSED_ERROR,
+            EVENT_STREAM_RPC_UNEXPECTED_ERROR,
+            EVENT_STREAM_RPC_CRT_ERROR
+        };
+
+        struct EventStreamRpcStatus
```
DavidOgunsAWS, line 253: This is too broad of a name and makes it unclear. EventStreamRpcResult is a suggestion

DavidOgunsAWS (changes requested):

eventstreamrpc/source/EventStreamClient.cpp
```diff
@@ -0,0 +1,1226 @@
...
+            }
+            else
+            {
+                responseFactory = m_responseRetriever.GetStreamingResponseFromModelName(GetModelName());
+            }
+
+            if (responseFactory == nullptr)
+            {
+                return EVENT_STREAM_RPC_UNMAPPED_DATA;
+            }
```
DavidOgunsAWS, line 986: The control flow for handling incoming data should defer to the Smithy model (through generated code+interfaces). This flow assumes that if it got a second message it must be a streaming operation, then if it can't find a handler for it, it it received unmapped data. This is confusing and should be avoided.

ipc/tests/GGCoreIpcClientTest.cpp
```diff
@@ -0,0 +1,75 @@
...
+            auto response = operation.GetResponse();
+            response.wait();
+        }
+
+        /* Publish to Topic */
+        {
+            Ipc::PublishToTopicOperation operation = client.NewPublishToTopic();
+            Ipc::PublishMessage publishMessage(allocator);
+            Ipc::PublishToTopicRequest request(
+                Aws::Crt::String("example"), Ipc::PublishMessage(publishMessage), allocator);
```
DavidOgunsAWS, line 62: Redundant copy of `Ipc::PublishMessage publishMessage`?

ipc/source/GGCoreIpcClient.cpp
```diff
@@ -0,0 +1,508 @@
...
+#include <sstream>
+
+namespace Aws
+{
+    namespace Eventstreamrpc
+    {
+        namespace Ipc
+        {
+            BinaryMessage::BinaryMessage(
+                Crt::Optional<Crt::Vector<uint8_t>> &&message,
```
DavidOgunsAWS, line 13: Remove field arguments from the constructor. If you have them there and additional fields are added, it will trivially break existing code when the order of arguments is changed or more are added.

If you want a convenient way to build/initialize objects in a fluent combined statement, and keep a reasonable path for backwards compatibility, you need to generate builders.

DavidOgunsAWS, line 13: Remove field arguments from the constructor. If you have them there and additional fields are added, it will trivially break existing code when the order of arguments is changed or more are added.

If you want a convenient way to build/initialize objects in a fluent combined statement and maintain backwards compatibility, you need to generate builders.

ipc/source/GGCoreIpcClient.cpp
```diff
@@ -0,0 +1,508 @@
...
+                authTokenPayloadSS << "{\"authToken\":\"" << finalAuthToken << "\"}";
+
+                if (!m_clientBootstrap)
+                {
+                    initializationPromise.set_value({EVENT_STREAM_RPC_INITIALIZATION_ERROR, 0});
+                    return initializationPromise.get_future();
+                }
+
+                Crt::Io::SocketOptions socketOptions;
+                socketOptions.SetSocketDomain(Crt::Io::SocketDomain::Local);
```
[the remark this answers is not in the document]
yourslab: It might be a good idea to have a Config class so that we won't need to introduce API changes if connect requires a different set of parameters in the future

yourslab:
Closing because #287 contains the generated code

PR #287, opened 2021-06-18
GG IPC & Echo Client

yourslab:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,8 @@ add_subdirectory(jobs)
 add_subdirectory(shadow)
 add_subdirectory(discovery)
 add_subdirectory(identity)
+add_subdirectory(eventstream_rpc)
+add_subdirectory(greengrass_ipc)
 if (NOT BYO_CRYPTO)
     # TODO: get these working with BYO_CRYPTO
     add_subdirectory(iotdevicecommon)
--- a/eventstream_rpc/source/EventStreamClient.cpp
+++ b/eventstream_rpc/source/EventStreamClient.cpp
@@ -1068,7 +1068,7 @@ namespace Aws
             Crt::Allocator *allocator) noexcept
             : m_operationModelContext(operationModelContext), m_messageCount(0), m_allocator(allocator),
               m_streamHandler(streamHandler), m_clientContinuation(connection.NewStream(*this)),
-              m_closeState(WONT_CLOSE), m_numCloses(0), m_streamClosedCalled(false)
+              m_expectedCloses(0), m_streamClosedCalled(false)
         {
         }
 
@@ -1083,10 +1083,8 @@ namespace Aws
         ClientOperation::~ClientOperation() noexcept
         {
             Close().wait();
-            if (m_numCloses.load() > 0)
-            {
-                m_closedPromise.get_future().wait();
-            }
+            std::unique_lock<std::mutex> lock(m_continuationMutex);
+            m_closeReady.wait(lock, [this]{return m_expectedCloses.load() == 0;});
         }
 
         TaggedResult::TaggedResult(Crt::ScopedResource<AbstractShapeBase> operationResponse) noexcept
@@ -1326,8 +1324,7 @@ namespace Aws
             if (messageFlags & AWS_EVENT_STREAM_RPC_MESSAGE_FLAG_TERMINATE_STREAM)
             {
                 const std::lock_guard<std::mutex> lock(m_continuationMutex);
-                m_closeState = WILL_CLOSE;
-                m_numCloses.fetch_add(1);
+                m_expectedCloses.fetch_add(1);
             }
 
             m_messageCount += 1;
@@ -1434,12 +1431,10 @@ namespace Aws
             /* Promises must be reset in case the client would like to send a subsequent request with the same
              * `ClientOperation`. */
             m_initialResponsePromise = {};
-            m_closedPromise = {};
+            //m_closedPromise = {};
             {
                 const std::lock_guard<std::mutex> lock(m_continuationMutex);
                 m_resultReceived = false;
-                if (m_closeState != WILL_CLOSE)
-                    m_closeState = WONT_CLOSE;
             }
 
             Crt::List<EventStreamHeader> headers;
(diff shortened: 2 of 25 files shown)


eventstreamrpc/source/EventStreamClient.cpp
```diff
@@ -0,0 +1,1301 @@
...
+        std::future<RpcError> ClientOperation::Close(OnMessageFlushCallback onMessageFlushCallback) noexcept
+        {
+            if (m_isClosed.load())
+            {
+                std::promise<RpcError> alreadyClosedPromise;
+                alreadyClosedPromise.set_value({EVENT_STREAM_RPC_STREAM_CLOSED_PREMATURELY, 0});
+                return alreadyClosedPromise.get_future();
+            }
+            else
+            {
```
DavidOgunsAWS, line 1313: The protection against closing twice is already accomplished via the underlying continuation. Just use it to send the necessary close message and close of the continuation. If it's already closed, it'll surface via the SendMessage() response

eventstreamrpc/tests/EventStreamClientTest.cpp
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        /* Rejected client-name header. */
+        {
+            TestLifecycleHandler lifecycleHandler;
+            ClientConnection connection(allocator);
+            connectionAmendment.AddHeader(EventStreamHeader(
+                Aws::Crt::String("client-name"), Aws::Crt::String("rejected.testy_mc_testerson"), allocator));
+            auto future = connection.Connect(config, &lifecycleHandler, messageAmender);
+            ASSERT_TRUE(future.get().baseStatus == EVENT_STREAM_RPC_CONNECTION_CLOSED_BEFORE_CONNACK);
```
DavidOgunsAWS, line 121: Should be ACCESS_DENIED

eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -0,0 +1,659 @@
...
+            /**
+             * Invoked when the continuation is closed.
+             *
+             * Once the continuation is closed, no more messages may be sent or received.
+             * The continuation is closed when a message is sent or received with
+             * the TERMINATE_STREAM flag, or when the connection shuts down.
+             */
+            void OnContinuationClosed() override;
+
+            const EventStreamHeader *GetHeaderByName(
```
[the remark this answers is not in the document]
yourslab: I would prefer this too but we can't wrap a reference with `Crt::Optional` and don't think we have any other way of returning that the header does not exist besides returning `nullptr`. We could make an output parameter wrapped with `Crt::Optional` but that would make it pass by copy. This method is also only used privately.

CMakeLists.txt
```diff
@@ -62,6 +62,8 @@ add_subdirectory(jobs)
 add_subdirectory(shadow)
 add_subdirectory(discovery)
 add_subdirectory(identity)
+add_subdirectory(eventstreamrpc)
+add_subdirectory(ipc)
```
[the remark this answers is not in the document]
yourslab: Sure, I prefer this too but was wondering whether it would be weird to prefix a directory with greengrass even though discovery is not.

eventstreamrpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -0,0 +1,659 @@
...
+        struct RpcError
+        {
+            EventStreamRpcStatusCode baseStatus;
+            int crtError;
+            operator bool() const noexcept { return baseStatus == EVENT_STREAM_RPC_SUCCESS; }
+        };
+
+        template <class T> class ProtectedPromise
+        {
+            /* A wrapper around std::promise so that it cannot be set twice without having to catch exceptions. */
```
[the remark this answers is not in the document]
yourslab: Sure, I can make it return a bool. Also, I actually try to avoid usage of this class unless absolutely necessary. However, some promises like the one used for Connect ACK or the initial response can be set twice because the connection or continuation can close at any time while data is being processed, so there's a potential race condition there. We could also bundle the mutex with the connection or continuation but that's slightly more repetitive.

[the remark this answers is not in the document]
DavidOgunsAWS: This race condition is the specific responsibility of the connection class. The goal is for it to be thread safe and manage it's own state correctly when it comes to concurrent operations.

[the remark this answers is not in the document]
yourslab: This is resolved in the latest iteration

eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -487,19 +517,32 @@ namespace Aws
             return onFlushPromise.get_future();
         }
 
-        std::future<RpcError> ClientConnection::Close() noexcept
+        void ClientConnection::Close() noexcept
         {
+            m_stateMutex.lock();
```
DavidOgunsAWS, line 461: The pattern here is more easily and safely done via RAII using: https://en.cppreference.com/w/cpp/thread/lock_guard

greengrass_ipc/source/platform/PosixConnectionConfig.cpp
```diff
@@ -0,0 +1,33 @@
+#include <aws/greengrass/GreengrassCoreIpcClient.h>
```
DavidOgunsAWS, line 1: Keep this version of the file, remove the Window's version and suppress the warning.

Likely we will want to have an aws_getenv() function added to handle whatever the best behavior per platform should be.

DavidOgunsAWS (approved):

greengrass_ipc/source/platform/PosixConnectionConfig.cpp
yourslab: Okay, I'll say that this is something we can do later on as introducing this may require more work. For now, we can suppress the warning in MSVC:
```
#if defined(_MSC_VER)
    #pragma warning(disable: 4996)
#endif

File: eventstream_rpc/tests/EventStreamClientTest.cpp [hunk 8]
```diff
@@ -504,7 +505,17 @@ static int s_TestStressClient(struct aws_allocator *allocator, void *ctx)
             messageData.SetStringMessage(expectedMessage);
             echoMessageRequest.SetMessage(messageData);
             auto requestFuture = echoMessage.Activate(echoMessageRequest, s_onMessageFlush);
-            requestFuture.wait();
+            std::future_status status = requestFuture.wait_for(std::chrono::seconds(1));
```
DavidOgunsAWS, line 508: Why is this waiting for a second?

eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -1434,12 +1431,10 @@ namespace Aws
             /* Promises must be reset in case the client would like to send a subsequent request with the same
              * `ClientOperation`. */
             m_initialResponsePromise = {};
-            m_closedPromise = {};
+            //m_closedPromise = {};
```
DavidOgunsAWS, line 1434: just delete

eventstream_rpc/tests/EventStreamClientTest.cpp
yourslab: This is removed now

eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -213,9 +213,9 @@ namespace Aws
             }
         }
 
-        bool ConnectionLifecycleHandler::OnErrorCallback(int errorCode)
+        bool ConnectionLifecycleHandler::OnErrorCallback(RpcError error)
         {
-            (void)errorCode;
+            (void)error;
             /* Returning true implies that the connection will close upon receiving an error. */
```
DavidOgunsAWS, line 219: Missed this before, but the documentation should read "the connection _should_ close as a result of encountering this error"

PR #305, opened 2021-07-27
Updated DeviceDefender task interface changes

DavidOgunsAWS:
*Issue #, if available:*

*Description of changes:*
* These changes are downstream of these changes: https://github.com/awslabs/aws-c-iot/pull/48 and remain in draft until it is pushed and given a pre-release tag
* The interface to start a DeviceDefender task has been updated for better safety and extensibility to include custom metrics. This device SDK needed updates to adhere to the new interface
* No functionality for custom metrics has been added yet

### DIFF
--- a/devicedefender/source/DeviceDefender.cpp
+++ b/devicedefender/source/DeviceDefender.cpp
@@ -131,6 +131,7 @@ namespace Aws
             {
                 aws_iotdevice_defender_task_clean_up(this->m_owningTask);
                 this->m_owningTask = nullptr;
+                m_status = ReportTaskStatus::Stopped;
             }
         }
 
--- a/devicedefender/source/DeviceDefender.cpp
+++ b/devicedefender/source/DeviceDefender.cpp
@@ -41,6 +41,7 @@ namespace Aws
             : OnTaskCancelled(std::move(onCancelled)), cancellationUserdata(cancellationUserdata),
               m_allocator(allocator), m_status(ReportTaskStatus::Ready),
               m_taskConfig{nullptr},
+              m_owningTask{nullptr},
               m_lastError(0),
               m_mqttConnection{mqttConnection},
               m_eventLoopGroup{eventLoopGroup}
@@ -116,7 +117,7 @@ namespace Aws
                     this->m_status = ReportTaskStatus::Running;
                 }
             }
-            return AWS_OP_SUCCESS;
+            return AWS_OP_ERR;
         }
 
         void ReportTask::StopTask() noexcept
@@ -131,9 +132,12 @@ namespace Aws
         ReportTask::~ReportTask()
         {
             StopTask();
-            aws_iotdevice_defender_config_clean_up(m_taskConfig);
+            if (m_taskConfig)
+            {
+                aws_iotdevice_defender_config_clean_up(m_taskConfig);
+                this->m_taskConfig = nullptr;
+            }
             this->m_owningTask = nullptr;
-            this->m_taskConfig = nullptr;
             this->m_allocator = nullptr;
             this->OnTaskCancelled = nullptr;
             this->cancellationUserdata = nullptr;
--- a/devicedefender/source/DeviceDefender.cpp
+++ b/devicedefender/source/DeviceDefender.cpp
@@ -103,6 +103,7 @@ namespace Aws
 
         int ReportTask::StartTask() noexcept
         {
+            int return_code = AWS_OP_ERR;
             if (m_taskConfig != nullptr && !m_lastError &&
                 (this->GetStatus() == ReportTaskStatus::Ready || this->GetStatus() == ReportTaskStatus::Stopped))
             {
@@ -114,14 +115,14 @@ namespace Aws
                 {
                     this->m_lastError = aws_last_error();
                     aws_raise_error(this->m_lastError);
-                    return AWS_OP_ERR;
                 }
                 else
                 {
                     this->m_status = ReportTaskStatus::Running;
+                    return_code = AWS_OP_SUCCESS;
                 }
             }
-            return AWS_OP_ERR;
+            return return_code;
         }
 
         void ReportTask::StopTask() noexcept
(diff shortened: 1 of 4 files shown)


devicedefender/source/DeviceDefender.cpp
```diff
@@ -37,34 +39,37 @@ namespace Aws
...
+            {
+                m_lastError = aws_last_error();
+            }
         }
 
         ReportTask::ReportTask(ReportTask &&toMove) noexcept
             : OnTaskCancelled(std::move(toMove.OnTaskCancelled)), cancellationUserdata(toMove.cancellationUserdata),
               m_allocator(toMove.m_allocator), m_status(toMove.m_status), m_taskConfig(std::move(toMove.m_taskConfig)),
-              m_owningTask(toMove.m_owningTask), m_lastError(toMove.m_lastError)
+              m_owningTask(toMove.m_owningTask), m_lastError(toMove.m_lastError),
```
bretambrose, line 66: Is this constructor necessary?

devicedefender/source/DeviceDefender.cpp
```diff
@@ -37,34 +39,37 @@ namespace Aws
...
+              m_mqttConnection(toMove.m_mqttConnection), m_eventLoopGroup(toMove.m_eventLoopGroup)
         {
-            m_taskConfig.cancellation_userdata = this;
-            toMove.OnTaskCancelled = nullptr;
-            toMove.cancellationUserdata = nullptr;
-            toMove.m_allocator = nullptr;
+            toMove.m_taskConfig = m_taskConfig;
+            toMove.m_owningTask = m_owningTask;
+            m_owningTask = nullptr;
```
bretambrose, line 71: This doesn't look right to me.  Why are we pushing things back into the thing we're moving out of?

For example here are several side-affect sequences, neither of which seem correct:

m_taskConfig(std::move(toMove.m_taskConfig))
toMove.m_taskConfig = m_taskConfig;

m_owningTask(toMove.m_owningTask)  // copy
toMove.m_owningTask = m_owningTask // does nothing
m_owningTask = nullptr; // why?
toMove.m_owningTask = nullptr;

devicedefender/source/DeviceDefender.cpp
```diff
@@ -37,34 +39,37 @@ namespace Aws
...
-                           this},
-              m_lastError(0)
+              m_allocator(allocator), m_status(ReportTaskStatus::Ready), m_taskConfig{nullptr}, m_owningTask{nullptr},
+              m_lastError(0), m_mqttConnection{mqttConnection}, m_eventLoopGroup(eventLoopGroup)
         {
+            (void)networkConnectionSamplePeriodSeconds;
+            struct aws_byte_cursor thingNameCursor = Crt::ByteCursorFromString(thingName);
+            m_lastError =
+                aws_iotdevice_defender_config_create(&m_taskConfig, allocator, &thingNameCursor, reportFormat);
+            if (AWS_OP_SUCCESS == m_lastError)
```
graebm, line 49: trivial: it's a little weird to save the `AWS_OP_` result in the same variable that's supposed to hold an `AWS_ERROR_`. They're conceptually two different types, even if they both happen to be ints 

consider either
```
if (aws_iotdevice_defender_config_create(...) == AWS_OP_SUCCESS)
```
or
```
int result = aws_iotdevice_defender_config_create(...);
if (result == AWS_OP_SUCESS)
    ...
```
```

File: devicedefender/source/DeviceDefender.cpp (hunk 1 above)
Comment (graebm) [inline devicedefender/source/DeviceDefender.cpp:66]: Are you sure you want to support move operations? It's a lot of subtle stuff that can accidentally go wrong. And if we ever need to pass a pointer to the C++ class down into C as user_data then then move is illegal

File: devicedefender/source/DeviceDefender.cpp [hunk 4]
```diff
@@ -100,37 +103,46 @@ namespace Aws
...
-
-                if (this->m_owningTask == nullptr)
+                if (AWS_OP_SUCCESS != aws_iotdevice_defender_task_create(
+                                          &m_owningTask,
+                                          this->m_taskConfig,
+                                          m_mqttConnection->GetUnderlyingConnection(),
+                                          aws_event_loop_group_get_next_loop(m_eventLoopGroup.GetUnderlyingHandle())))
                 {
                     this->m_lastError = aws_last_error();
                     aws_raise_error(this->m_lastError);
```
graebm, line 117: trivial: there's no point in re-raising the error

devicedefender/source/DeviceDefender.cpp
```diff
@@ -100,37 +103,46 @@ namespace Aws
...
-                    return AWS_OP_ERR;
                 }
                 else
                 {
                     this->m_status = ReportTaskStatus::Running;
+                    return_code = AWS_OP_SUCCESS;
                 }
             }
-            return AWS_OP_SUCCESS;
+            return return_code;
```
graebm, line 125: With this change, if the first `if` statement fails, we'll end up returning AWS_OP_ERR without having ever raised an error

graebm (approved):
fix & ship

devicedefender/source/DeviceDefender.cpp
DavidOgunsAWS: I'm not sure. It was already there so I made adjustments as necessary rather than figure out if it was safe to delete.

DavidOgunsAWS: I didn't, but after evaluating that callback_userdata is given a value of `this`, it's clear that a move after that won't be valid. This class has to be non-copyable, non-moveable. Making the adjustments

devicedefender/source/DeviceDefender.cpp
DavidOgunsAWS: Deleted this entire function

devicedefender/source/DeviceDefender.cpp
DavidOgunsAWS: Removed

devicedefender/source/DeviceDefender.cpp
DavidOgunsAWS: True. Raising AWS_ERROR_INVALID_STATE in this situation

PR #316, opened 2021-09-06
rename topic and client id of basic_pub_sub example

mcerdeiro:
Rename the default topic and client id to match
the naming from other examples like the from python and java sdk.

*Issue #, if available:*

Background: One common way to use this SDK as newbie in AWS IoT is to use the Onboard -> Get started "tutorial". I used this option for Linux/Python. This created automatically a working example sending and receiving messages in python. The next step was to reproduce the same with C++ SDK. Trying to reproduce the same behavior in C++ is for someone new in the technology complicated due to 2 small differences addressed in this pr. I also saw multiple people having this similar issue googling for the error message:

> Connection failed with error libaws-c-mqtt: AWS_ERROR_MQTT_UNEXPECTED_HANGUP, The connection was closed unexpectedly.

This PR "as is" makes only sense if also the default policy created during "Onboard" -> Get started is extended adding: 

arn:aws:iot:<zone>:<id>:topicfilter/sdk/test/cpp"

if not the PR can be also adapted to use the topic sdk/test/Python or sdk/test/java.

*Description of changes:*

Rename topic and client id to match the naming of the policies created for python example using the Onboard -> Get started menu.

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -61,7 +61,7 @@ and receive.
         "iot:Receive"
       ],
       "Resource": [
-        "arn:aws:iot:<b>region</b>:<b>account</b>:topic/test/topic"
+        "arn:aws:iot:<b>region</b>:<b>account</b>:topic/sdk/test/cpp"
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


jmklix:
This is not something what we want to change to be specific for each language sdk

PR #350, opened 2021-11-17
Moved and renamed tunnel notification sample and added a secure tunne…

sbSteveK:


samples/secure_tunneling/secure_tunnel/main.cpp
```diff
@@ -0,0 +1,278 @@
...
+            .WithOnSessionReset(OnSessionReset)
+            .Build();
+
+    if (!mSecureTunnel)
+    {
+        fprintf(stderr, "Secure Tunnel Creation failed");
+        exit(-1);
+    }
+
+    mSecureTunnel->Connect();
```
graebm, line 256: check Connect() for failure like
```
if (mSecureTunnel->Connect() == AWS_OP_ERR) {
    fprintf(stderr, "Secure Tunnel Connect call failed: %s\n", ErrorDebugString(LastError()));
    exit(-1);
}
```

samples/secure_tunneling/secure_tunnel/main.cpp
```diff
@@ -0,0 +1,278 @@
...
+    int messageCount = 0;
+
+    while (1)
+    {
+        std::this_thread::sleep_for(3000ms);
+        if (localProxyMode == AWS_SECURE_TUNNELING_SOURCE_MODE)
+        {
+            messageCount++;
+            string toSend = to_string(messageCount) + ": " + message.c_str();
```
graebm, line 265: So ... `mSecureTunnel->Connect()` that gets called above is async. Even if it returns AWS_OP_SUCCESS, the connection still hasn't succeeded, it just means the connection attempt successfully *started*.

You need to wait until the connection is successful before trying to send data. This is succeeding right now because it waits 3 seconds before sending data, and you happen to connect in those 3 seconds, but that won't always be the case..

You probably want to do something like the MQTT samples, that use synchronization primitives to halt the main thread until the connection callback fires. Look at how [connectionCompletedPromise](https://github.com/aws/aws-iot-device-sdk-cpp-v2/blob/2d8447ccb2c4a4a89b504a382c5329c8181140a1/samples/mqtt/basic_pub_sub/main.cpp#L434) is used, the main thread waits on `connectionCompletedPromise` to know when the connection attempt completes, and the callbacks set `connectionCompletedPromise` to indicate whether it succeeded or failed

samples/secure_tunneling/secure_tunnel/main.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+            if (!mSecureTunnel->SendData(ByteCursorFromCString(toSend.c_str())))
+            {
+                fprintf(stdout, "Sending Message:\"%s\"\n", toSend.c_str());
+                if (messageCount >= 5)
+                {
+                    fprintf(stdout, "Closing Connection\n");
+                    mSecureTunnel->Close();
```
graebm, line 273: like Connect(), the Close() function can also fail. probably want to print the error and exit if it fails

graebm: Also, like Connect() the Close() call is async. Probably want to use another `std::promise` so that you wait for the connection closing to complete before exiting the program (currently, you're still looping every 3 seconds, it's just that the connection *tends to* finish before that.

samples/README.md
```diff
@@ -443,15 +443,45 @@ using a permanent certificate set, replace the paths specified in the `--cert` a
         --csr /tmp/deviceCert.csr
 ```
 
-## Secure Tunneling
+## Secure Tunnel Notification
```
graebm, line 446: trivial: maybe explain the "Secure Tunnel" sample before the "Secure Tunnel Notification" sample, since the Notification one is a little more niche

samples/secure_tunneling/secure_tunnel/main.cpp
```diff
@@ -0,0 +1,278 @@
...
+                else
+                {
+                    fprintf(stderr, "Send Data Failed with error code %d\n", error_code);
+                }
+                break;
+        }
+    };
+
+    auto OnDataReceive = [&](const struct aws_byte_buf &data) {
+        receivedData = std::string((char *)data.buffer, data.len);
```
graebm, line 205: A little bit weird to have declared `receivedData` outside this function. Just make it a local variable

I think I already said that...

samples/secure_tunneling/secure_tunnel/main.cpp
```diff
@@ -0,0 +1,278 @@
...
+int main(int argc, char *argv[])
+{
+    ApiHandle apiHandle;
+
+    String region;
+    String endpoint;
+    String caFile;
+    String accessToken;
+    String message = "Hello World";
+    string receivedData = "";
```
graebm, line 65: declaring `string receivedData` up here at the top of main() is weird.
It's only used in the OnDataReceived() callback. I vote you just move it in there

samples/secure_tunneling/secure_tunnel/main.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+int main(int argc, char *argv[])
+{
+    ApiHandle apiHandle;
+
+    String region;
+    String endpoint;
+    String caFile;
+    String accessToken;
+    String message = "Hello World";
```
graebm, line 64: trivial: weird mix of `Aws::Crt::String` and `std::string`
maybe just stick with std::string because this is a sample, and they shouldn't be using our weird internal typedefs

samples/secure_tunneling/secure_tunnel/main.cpp
```diff
@@ -0,0 +1,278 @@
...
+    string receivedData = "";
+    aws_secure_tunneling_local_proxy_mode localProxyMode;
+    std::shared_ptr<SecureTunnel> mSecureTunnel;
```
graebm, line 67: trivial: a little weird to prefix this one variable with "m"
in c++ land that prefix often denotes that it's a member of a class
I suggest you just go with `secureTunnel`

bretambrose (commented):
At the risk of being a jerk I feel like the proxy and non-proxy samples share so much code that they could be combined.

PR #369, opened 2022-01-05
Greengrass IPC API updates

bretambrose:
New approach to GG IPC operation management.

* Make ClientContinuationHandler's destructor virtual so that it can actually be used as a pointed-to value
* ClientOperation uses public inheritance so that we can use the virtual destructor
* Most importantly, a new set of per-operation APIs that return a pointer to the operation, not the operation's value itself
* The new pointer-returning APIs that use a stream handler take the stream handler by shared ptr to guarantee the handler lives sufficiently long 

### DIFF
--- a/eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
+++ b/eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
@@ -292,7 +292,7 @@ namespace Aws
              * the TERMINATE_STREAM flag, or when the connection shuts down.
              */
             virtual void OnContinuationClosed() = 0;
-            ~ClientContinuationHandler() noexcept;
+            virtual ~ClientContinuationHandler() noexcept;
 
           private:
             friend class ClientContinuation;
@@ -487,7 +487,7 @@ namespace Aws
             const ServiceModel &m_serviceModel;
         };
 
-        class AWS_EVENTSTREAMRPC_API ClientOperation : private ClientContinuationHandler
+        class AWS_EVENTSTREAMRPC_API ClientOperation : public ClientContinuationHandler
         {
           public:
             ClientOperation(
--- a/eventstream_rpc/tests/EchoTestRpcClient.cpp
+++ b/eventstream_rpc/tests/EchoTestRpcClient.cpp
@@ -1,3 +1,4 @@
+#include <aws/crt/Types.h>
 #include <aws/crt/io/Bootstrap.h>
 #include <awstest/EchoTestRpcClient.h>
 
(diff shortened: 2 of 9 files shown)


graebm (commented):
beautiful

clogwog:
thank you for looking into this. The Pointer returned from the new `NewPtrSubscribeToIoTCore` will solve the problem. 
Is there a specific way to dispose of the `std::unique_ptr<SubscribeToIoTCoreOperation>` whenever it isn't needed anymore ? Once this has been approved and merged please let me know and i'm willing to make an update to the examples in the documentation if you like. 
thanks again ! (i started using  `static SubscribeToIoTCoreOperation operation = ipcClient.NewSubscribeToIoTCore(streamHandler);` just to be able to keep going..but it isn't very elegant. this is much better)

bretambrose:
>  Is there a specific way to dispose of the `std::unique_ptr<SubscribeToIoTCoreOperation>` whenever it isn't needed anymore ? 

You can assign it to null or let it drop out of scope from wherever you've seated it and it will get destroyed naturally.

eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -487,7 +487,7 @@ namespace Aws
             const ServiceModel &m_serviceModel;
         };
 
-        class AWS_EVENTSTREAMRPC_API ClientOperation : private ClientContinuationHandler
+        class AWS_EVENTSTREAMRPC_API ClientOperation : public ClientContinuationHandler
```
graebm, line 490: trivial: did the inheritance really need to change from private to public?

eventstream_rpc/tests/EchoTestRpcClient.cpp
```diff
@@ -28,11 +29,25 @@ namespace Awstest
         return GetAllProductsOperation(
             m_connection, m_echoTestRpcServiceModel.m_getAllProductsOperationContext, m_allocator);
     }
+
+    std::unique_ptr<GetAllProductsOperation> EchoTestRpcClient::NewPtrGetAllProducts() noexcept
+    {
+        return std::unique_ptr<GetAllProductsOperation>(Aws::Crt::New<GetAllProductsOperation>(
+            m_allocator, m_connection, m_echoTestRpcServiceModel.m_getAllProductsOperationContext, m_allocator));
```
graebm, line 36: question: Is there any way the constructor can create an "invalid" version of the Operation, where the user needs to do like:
```
auto operationPtr = client->NewPtrGetAllProducts();
if (!operationPtr) {
    blugh
}
if (!(*operationPtr)) {
    blugh
}
```

If so, I'd strongly prefer we return an empty unique_ptr, instead of returning a unique_ptr that has something in it, but that thing's invalid.

If it's impossible for the custructor to return an invalid thing, then nevermind this is fine

graebm (approved):

eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
bretambrose: Yes, otherwise the destructor was uncallable.

eventstream_rpc/tests/EchoTestRpcClient.cpp
bretambrose: It doesn't look like the constructor (in particular, ClientOperation's constructor can enter an invalid state short of allocation failure).

clogwog:
feedback: re-installed library and changed the code to use the pointer and everything is working ! 

`       // static SubscribeToIoTCoreOperation operation = ipcClient.NewSubscribeToIoTCore(streamHandler);
        // TODO Change to the pointer once iot sdk c++ supports this

        std::shared_ptr<SubscribeToIoTCoreStreamHandler> pstreamhandler = std::shared_ptr<SubscribeToIoTCoreStreamHandler>(&streamHandler);
        p_operation = ipcClient.NewPtrSubscribeToIoTCore(pstreamhandler);
`

thank you !

bretambrose:
You probably want to allocate your stream handler rather than do what you've written there which will crash with a double delete, one from the shared_ptr and one from the local object.

clogwog:
@bretambrose 
does the usage for a local (non iot-core) subscription look like:

```
      if (! my_ipc_data->ipcClient.IsConnected())
      {
        Aws::Eventstreamrpc::RpcError connectionStatus = my_ipc_data->ipcClient.Connect(self->ipcLifecycleHandler).get();
        if (connectionStatus != Aws::Eventstreamrpc::EventStreamRpcStatusCode::EVENT_STREAM_RPC_SUCCESS)
          return false;
      }

      my_ipc_data->subscription_request.SetTopic(my_ipc_data->topic);
      my_ipc_data->p_streamhandler = std::shared_ptr<SubscribeInsertMetaDataResponseHandler>(new SubscribeInsertMetaDataResponseHandler());
      my_ipc_data->p_operation = my_ipc_data->ipcClient.NewPtrSubscribeToTopic(my_ipc_data->p_streamhandler);
      auto activate = my_ipc_data->p_operation->Activate(my_ipc_data->subscription_request, nullptr);
      activate.wait();
```

i'm asking because i'm finding that
`      my_ipc_data->p_operation = my_ipc_data->ipcClient.NewPtrSubscribeToTopic(my_ipc_data->p_streamhandler);`
never returns ?

PR #377, opened 2022-02-08
Refactor of GG IPC public API

bretambrose:
**Breaking change**

* Removes the local-object-returning IPC APIs in factor of shared_ptr management
* Refactors eventstream rpc objects to use shared_ptr for the stream handler, avoiding destruction order crashes


TwistedTwigleg (approved):
Looks good to me 👍 
Posted a question on Slack about it, but not a blocking thing just a curiosity question.

PR #388, opened 2022-02-18
Refactor of command line argument parsing to single file

TwistedTwigleg:
*Description of changes:*

Refactors command line parsing into a single file shared across all samples. This reduces the duplicate code, makes the samples smaller, and also has the benefit of additions made to the command line parsing file benefiting all samples.

This PR modifies all of the samples to use this shared file, modifies some wording for better consistency across samples, and adjusts the secure_tunnel sample to use the `Aws::String` class instead of `std::string`.

_____________

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*

### DIFF
--- a/samples/greengrass/basic_discovery/main.cpp
+++ b/samples/greengrass/basic_discovery/main.cpp
@@ -42,17 +42,14 @@ int main(int argc, char *argv[])
     Utils::CommandLineUtils cmdUtils = Utils::CommandLineUtils();
     cmdUtils.RegisterProgramName("basic-discovery");
     cmdUtils.AddCommonMQTTCommands();
+    cmdUtils.AddCommonProxyCommands();
+    cmdUtils.AddCommonTopicMessageCommands();
     cmdUtils.RemoveCommand("endpoint");
     cmdUtils.RegisterCommand(
         "region", "<str>", "The region for your Greengrass groups (optional, default='us-east-1').");
     cmdUtils.RegisterCommand("thing_name", "<str>", "The name of your IOT thing");
-    cmdUtils.RegisterCommand("topic", "<str>", "Targeted topic (optional, default='test/topic')");
     cmdUtils.RegisterCommand(
         "mode", "<str>", "Mode options: 'both', 'publish', or 'subscribe' (optional, default='both').");
-    cmdUtils.RegisterCommand("message", "<str>", "Message to publish (optional, default='Hello World').");
-    cmdUtils.RegisterCommand(
-        "proxy_host", "<str>", "Proxy host to use for discovery call. Default is to not use a proxy. (optional)");
-    cmdUtils.RegisterCommand("proxy_port", "<int>", "Proxy port to use for discovery call. (optional, default='0')");
     const char **const_argv = (const char **)argv;
     cmdUtils.SendArguments(const_argv, const_argv + argc);
 
--- a/samples/greengrass/basic_discovery/main.cpp
+++ b/samples/greengrass/basic_discovery/main.cpp
@@ -19,7 +19,7 @@
 using namespace Aws::Crt;
 using namespace Aws::Discovery;
 
-int main(const int argc, const char *argv[])
+int main(int argc, char *argv[])
 {
     /************************ Setup the Lib ****************************/
     /*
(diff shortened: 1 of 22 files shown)


TwistedTwigleg:
Note: Going to change things a little bit after doing the Python simplification. In Python, the arguments help have it in the format where the data in between `<>` shows the data type (examples: `<str>`, `<int>`, `<path>`), which I think makes a lot of sense and so I want to change the C++ code accordingly to follow the same format.

samples/utils/CommandLineUtils.h
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * A struct to hold the command line options that can be passed to the program from the terminal/console.
+     */
+    struct CommandLineOption
+    {
+        Aws::Crt::String CommandName;
+        Aws::Crt::String ExampleInput;
+        Aws::Crt::String HelpOutput;
+
+        CommandLineOption(Aws::Crt::String p_Name, Aws::Crt::String p_ExampleInput, Aws::Crt::String p_Help = "")
```
bretambrose, line 20: Let's not do "p_" on parameters.  Nothing wrong with it, but it's not a pattern that exists anywhere in the rest of the codebase.

samples/utils/CommandLineUtils.h
```diff
@@ -0,0 +1,139 @@
...
+#include <aws/crt/Types.h>
+
+namespace Utils
+{
+    /**
+     * A struct to hold the command line options that can be passed to the program from the terminal/console.
+     */
+    struct CommandLineOption
+    {
+        Aws::Crt::String CommandName;
```
bretambrose, line 16: It's not done with 100% consistency, but using camel case with a "m_" prefix for member names is usually what you'll see in the codebase, ie. m_commandName

samples/utils/CommandLineUtils.h
bretambrose, line 20: It doesn't matter at all performance-wise, but from a C++ parameter passing perspective, you're passing strings by value and generating unnecessary copies with a direct assignment.

Pre-C++-11 you'd always pass by const ref in this situation.  I think these days, the current "wisdom/guidance" from the C++ guru camp is pass by value but assign by move:

m_commandName = std::move(p_Name);

ostensibly in the name of supporting parameters that are rvals and certain compiler optimizations when passing small structs.

So while it doesn't matter much in this case, it's good to get in the practice/habit of evaluating the parameter ownership semantics and following the appropriate widely-accepted pattern.

samples/utils/CommandLineUtils.h
```diff
@@ -0,0 +1,139 @@
...
+         */
+        void UpdateCommandHelp(Aws::Crt::String CommandName, Aws::Crt::String NewCommandHelp);
+
+        /**
+         * Called to give the class a copy of the begin and end character pointers that contain the arguments from the
+         * terminal/console
+         * @param begin The beginning of terminal/console input
+         * @param end The end of terminal/console input
+         */
+        void SendArguments(char **begin, char **end);
```
bretambrose, line 76: I'd just reflect argv, argc here rather than the raw array end.  Also these should be const.

TwistedTwigleg:
Thanks for the review! I'll make the changes soon 👍

TwistedTwigleg:
Thanks for the review! I updated the code accordingly 👍

samples/greengrass/basic_discovery/main.cpp
```diff
@@ -19,7 +19,7 @@
 using namespace Aws::Crt;
 using namespace Aws::Discovery;
 
-int main(int argc, char *argv[])
+int main(const int argc, const char *argv[])
```
bretambrose, line 22: Sorry, I wasn't clear in the feedback.  The main()s were fine as is.  It was the function you made that took two char ** that should have been const (if it still exists).

TwistedTwigleg: Oh, my bad! I adjusted the code so it no longer sets `main` to const while still keeping the function that takes two char** (`SendArguments`) const.

samples/utils/CommandLineUtils.cpp
```diff
@@ -0,0 +1,174 @@
...
+            "ca_file", "<path>", "Path to AmazonRootCA1.pem (optional, system trust store used by default)."));
+    }
+
+    void CommandLineUtils::AddCommonProxyCommands()
+    {
+        RegisterCommand("proxy_host", "<str>", "Host name of the proxy server to connect through (optional)");
+        RegisterCommand("proxy_port", "<int>", "Port of the proxy server to connect through (optional, default='8080'");
+    }
+
+    void CommandLineUtils::AddCommonx509Commands()
```
bretambrose, line 138: I'd capitalize the X here.

bretambrose (approved):
Fix n ship

TwistedTwigleg:
Thanks! Will merge as soon as the CI is finished

PR #401, opened 2022-03-15
Device Defender custom metrics support

TwistedTwigleg:
*Description of changes:*

Adds support for registering custom metrics to the Device Defender service. Exposes functions to hook into Device Defender number, number-list, string-list, and ip-list metric gathering. Adds new functions to `Aws::Iotdevicedefenderv1` to register these custom metrics either by passing a function pointer or a `std::function`. Adds predefined custom metric functions that will get CPU usage (in percent), memory usage (in kilobytes) and processor count.

This PR also adds a device defender sample to show how to use and register custom metrics and custom metrics tests to the existing Device Defender tests.

_____________


devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -77,6 +77,148 @@ namespace Aws
...
+
+            /**
+             * Registers a custom metric number that will report the number of processors automatically into a custom
+             * metric called "processor_count". Calling this function will make the task report processor count each
+             * time a report is generated.
+             * @return AWS_OP_SUCCESS if the custom metric was registered successfully.
+             *      Will return AWS_OP_ERR if the CPU usage cannot be registered
+             */
+            int RegisterCustomMetricProcessorCount();
+
```
TwistedTwigleg, line 221: These are technically not needed, they are just helper functions that use the new custom metrics functionality to add support for tracking these metrics.

By putting them here they make these metrics easy to add, but I'm not sure if we want them here or if instead we should have it just in the sample and point users to the sample to see how to implement tracking for these features.

On one hand, removing them would reduce the code and keep it lean, but on the other, it would require the user to implement this commonly requested functionality themselves. It is not too hard to add

TwistedTwigleg:
*Note* This PR relies on the changes in https://github.com/awslabs/aws-c-iot/pull/67 in order to compile successfully, since it adds some new functions for custom metrics using doubles.

devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -77,6 +77,232 @@ namespace Aws
...
+             * "metricFunc".
+             * @param metricName The key name for the data.
+             * @param metricFunc The function that is called to get the number list data.
+             * @return AWS_OP_SUCCESS if the custom metric was registered successfully.
+             *          Will return AWS_OP_ERR if the number list cannot be registered or the data passed
+             *          ("metricName" or "metricFunc") is incorrect.
+             */
+            int RegisterCustomMetricNumberList(
+                aws_byte_cursor metricName,
+                std::function<int(aws_array_list *, void *)> *metricFunc);
```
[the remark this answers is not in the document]
graebm: I'd even be tempted to let the callback *return* a std::vector instead of modify one in-place

but I'm not 100% sure what the compiler would end up doing with that, if it would end up another needless copy or not. I don't understand [copy-ellision](https://en.cppreference.com/w/cpp/language/copy_elision) and return-value-optimization well enough

but in other languages I *would* let the callback just return the vector by value, because I know that wouldn't be an extra copy in other languages

TwistedTwigleg:
I pushed the first part of the changes based on the code review, but I have not fully implemented it just yet. Currently custom metrics for strings and IP lists are not working, causing segmentation faults when they are written to JSON. I think it is the strings are going out of reference.
Regardless, I wanted to push my work so I didn't lose anything but it's not quite ready yet.

TwistedTwigleg:
Thanks for the review! I adjusted the code and it's all working again 👍 
(Builds fail due to needing `aws-c-iot` changes)

devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -98,6 +173,57 @@ namespace Aws
                 void *cancellationUserdata = nullptr) noexcept;
 
             static void s_onDefenderV1TaskCancelled(void *userData);
+
+            struct customMetricData
+            {
+                size_t index;
+                ReportTask *task;
+            };
+            static int s_getCustomMetricNumber(double *output, void *customData);
```
bretambrose, line 182: Overall, this approach feels very C-like and I think it could be made a lot cleaner using C++ semantics.

My suggestion would be to:

1. Make a simple, flat class hierarchy: CustomMetricBase -> {CustomMetricNumber, CustomMetricNumberList, CustomMetricStringList, CustomMetricIpList}
2. Each leaf would hold a moved version of the passed in gathering function as well as a dynamically constructed function that we'll pass in to aws-c-iot that has captured the ReportTask context via lambda capture so there's no need for indexing or backpointers or anything like that.  See std::function::target for converting between std::function and C function pointer.
3. Have a single vector of CustomMetricBase *

devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -77,6 +77,81 @@ namespace Aws
...
+            void RegisterCustomMetricMemoryUsage();
+
+            /**
+             * Registers a custom metric number that will report the number of processes automatically into a custom
+             * metric called "process_count". Calling this function will make the task report processor count each
+             * time a report is generated.
+             */
+            void RegisterCustomMetricProcessCount();
+
+            std::function<int(double *)> *GetStoredCustomMetricNumber(size_t &index);
```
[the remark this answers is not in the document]
TwistedTwigleg: These functions are needed to get the stored custom metric function in the ReportTask from one of the `s_getCustomMetric*` statoc functions, as the vectors that store the custom metric functions are private and stored in the ReportTask reference.

I think changing the code to use more C++ semantics would remove the need for these functions though.

TwistedTwigleg:
Thanks for the review! I adjusted the code accordingly 👍 

One somewhat large change is how metrics are stored is mentioned below:

_______

(**Can be skipped** - just ranting about what I tried for my own reference)

I wasn't quite able to do the lambda thing with std function, as it seems getting a pointer to a C function from a C++ function and then passing that to C to be consumed is incredibly difficult and I could only find a single code reference where someone said they got it working. Unfortunately, it caused segmentation faults whenever I tried to run it and upon further reading in other sources, I'm not sure the snippet I found actually worked to begin with.

The issue stems from the problem that you cannot make a pointer to a C function without making a pointer to the function pointer (so `**function` rather than `*function`) if you want to assign it to a variable. The snippet I found that lead to a successful compile took the function pointer pointer (`**function`) and then cast it to a pointer to a function (`*pointer`) using: `*function = *(**function)` (the code wasn't that exactly, but that was the gist of it). It still caused a segmentation fault though no matter how I stored the pointers (including storing the std::function, and both pointers (`**function` and `*function`))
This code has a decent snippet showing, more or less, the code I was trying: [CPlusPlus.com](http://www.cplusplus.com/forum/general/63552/#msg343837)

I would have tried just getting a pointer to the function directly using `target<function>`, but the compiler *really* doesn't like it and from what I found online, it doesn't seem to be possible? I couldn't find any working examples of it at least and my experiments didn't yield any results.

I read more on it and concluded that it *probably* is possible, but with my current experience I have no idea how to implement it . The possible solutions I found were never quite what I was looking for, difficult to implement, and hard to follow. I decided to go for a different approach that I *think* is the best of both worlds. There was lots of calling C-style functions from C++, but not converting a std::function to a C-style function pointer to be invoked in C that I could find.

________

The new approach uses a vector of base classes, where the base class is really just a container for an allocator (needed for making strings). Then the other classes extend this base class, take a `std::function` of the proper type, and converts it to the necessary custom metric.

It is much cleaner than the previous implementation, requires only a single vector for storing all of the custom metrics, is easy to extend should we need to add different metrics in the future, and should hopefully also be easier to follow than the old C-like version.

devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -84,7 +200,9 @@ namespace Aws
              * @param metricName The key name for the data.
              * @param metricFunc The function that is called to get the number data.
              */
-            void RegisterCustomMetricNumber(const std::string &metricName, std::function<int(double *)> &metricFunc);
+            void RegisterCustomMetricNumber(
+                const Crt::String &metricName,
+                CustomMetricNumberFunction &metricFunc) noexcept;
```
bretambrose, line 205: I would make these rvalue references and move them internally

devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -31,6 +31,122 @@ namespace Aws
...
+         */
+        class AWS_IOTDEVICEDEFENDER_API CustomMetricStringList : public CustomMetricBase
+        {
+          public:
+            CustomMetricStringListFunction function;
+            void SetCustomMetricData(CustomMetricStringListFunction inputFunction, Crt::Allocator *inputAllocator);
+            static int GetMetricFunction(aws_array_list *output, void *data);
+
+            /*
+            void SetCustomMetricData(CustomMetricStringListFunction inputFunction, Crt::Allocator *inputAllocator)
```
bretambrose, line 97: Commented out impls can probably be removed

devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -174,56 +287,44 @@ namespace Aws
...
-            static int s_getCustomMetricIpList(aws_array_list *output, void *customData);
-            std::vector<std::function<int(double *)>> m_storedCustomMetricsNumberFunctions;
-            std::vector<std::function<int(std::vector<double> *)>> m_storedCustomMetricsNumberListFunctions;
-            std::vector<std::function<int(std::vector<std::string> *)>> m_storedCustomMetricsStringListFunctions;
-            std::vector<std::function<int(std::vector<std::string> *)>> m_storedCustomMetricsIpListFunctions;
-
-            std::vector<customMetricData *> storedCustomMetricData = std::vector<customMetricData *>();
+            // Holds all of the custom metrics created for this task. These are pointers that will be
+            // automatically created and cleaned by ReportTask when it is destroyed.
+            Crt::Vector<CustomMetricBase *> storedCustomMetrics;
```
bretambrose, line 292: Use shared_ptr or unique_ptr.  I'd use shared_ptr even if not necessary as then the deleter can be hidden from the type and you shouldn't need to keep an allocator member either.

devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -31,6 +31,122 @@ namespace Aws
...
+            Crt::Allocator *m_allocator;
+        };
+        /**
+         * A base class used to store all custom number metrics. Only used internally.
+         */
+        class AWS_IOTDEVICEDEFENDER_API CustomMetricNumber : public CustomMetricBase
+        {
+          public:
+            CustomMetricNumberFunction function;
+            void SetCustomMetricData(CustomMetricNumberFunction inputFunction, Crt::Allocator *inputAllocator);
```
bretambrose, line 54: Why aren't these SetCustomMetricData functions constructors?

devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -31,6 +31,122 @@ namespace Aws
...
+          public:
+            Crt::Allocator *m_allocator;
+        };
+        /**
+         * A base class used to store all custom number metrics. Only used internally.
+         */
+        class AWS_IOTDEVICEDEFENDER_API CustomMetricNumber : public CustomMetricBase
+        {
+          public:
+            CustomMetricNumberFunction function;
```
bretambrose, line 53: m_metricFunction

devicedefender/source/DeviceDefender.cpp
```diff
@@ -22,11 +22,81 @@ namespace Aws
...
+        {
+            function = std::move(inputFunction);
+            m_allocator = inputAllocator;
+        }
+        int CustomMetricNumberList::GetMetricFunction(aws_array_list *output, void *data)
+        {
+            CustomMetricNumberList *metric = (CustomMetricNumberList *)data;
+            Crt::Vector<double> function_data = Crt::Vector<double>();
+            int returnValue = metric->function(&function_data);
+            for (size_t i = 0; i < function_data.size(); i++)
```
bretambrose, line 52: consider std::for_each or similar for these copy loops

devicedefender/source/DeviceDefender.cpp
```diff
@@ -22,11 +22,81 @@ namespace Aws
...
+            {
+                aws_array_list_push_back(output, &function_data.at(i));
+            }
+            return returnValue;
+        }
+
+        // Custom string list metric setup and metric function getter.
+        void CustomMetricStringList::SetCustomMetricData(
+            CustomMetricStringListFunction inputFunction,
+            Crt::Allocator *inputAllocator)
```
bretambrose, line 62: It's not clear why inputAllocator is needed anywhere in these.

devicedefender/source/DeviceDefender.cpp
```diff
@@ -127,216 +191,90 @@ namespace Aws
                 aws_iotdevice_defender_config_clean_up(m_taskConfig);
                 this->m_taskConfig = nullptr;
             }
-            for (size_t i = 0; i < storedCustomMetricData.size(); i++)
+            for (size_t i = 0; i < storedCustomMetrics.size(); i++)
```
bretambrose, line 194: Once these are shared pointers you can remove this entirely

devicedefender/source/DeviceDefender.cpp
TwistedTwigleg: It is needed for making an `aws_string` in the GetMetricFunction for `StringList`s and `IPAddressList`s.

TwistedTwigleg:
Thanks for the review! I implemented the changes accordingly 👍 

One problem I have now though that I have not solved yet is now I am getting a memory leak of 5 allocations each around 48 bytes. This only started after I made the `shared_ptr` change, so I'm guessing it is related to that. My hunch is that it is using `delete` instead of `aws::delete` and that is leading to the issue. However, nothing I tried seemed to make much of a difference.
I am guessing I am missing something small, but I haven't so far been able to find it. Is there something I need to do with constructing/deconstructing the shared pointers? My guess is that's probably what is causing the leak.

TwistedTwigleg:
I fixed the memory leak issue. Turns out I just needed to use the normal new/delete with smart pointers (like what is done elsewhere in the code) and that fixed it. 😅

devicedefender/source/DeviceDefender.cpp
```diff
@@ -114,6 +190,189 @@ namespace Aws
...
+            aws_iotdevice_defender_config_register_number_list_metric(
+                m_taskConfig, &cursor, CustomMetricNumberList::GetMetricFunction, data.get());
+        }
+
+        void ReportTask::RegisterCustomMetricStringList(
+            const Crt::String &&metricName,
+            CustomMetricStringListFunction &&metricFunc) noexcept
+        {
+            std::shared_ptr<CustomMetricStringList> data =
+                std::shared_ptr<CustomMetricStringList>(new CustomMetricStringList(metricFunc, m_allocator));
```
bretambrose, line 222: when making shared_ptrs from CRT allocators, you'll want to use Aws::Crt::MakeShared

devicedefender/source/DeviceDefender.cpp
```diff
@@ -114,6 +190,189 @@ namespace Aws
...
+            uint64_t *totalSystem,
+            uint64_t *totalIdle)
+        {
+            *totalUser = 0;    // prevent warnings over unused parameter on Windows and Mac
+            *totalUserLow = 0; // prevent warnings over unused parameter on Windows and Mac
+            *totalSystem = 0;  // prevent warnings over unused parameter on Windows and Mac
+            *totalIdle = 0;    // prevent warnings over unused parameter on Windows and Mac
+
+// Get the CPU usage from Linux
+#if defined(__linux__) || defined(__unix__)
```
bretambrose, line 309: Ideally, this should be moved into an aws-c-common or aws-c-io API with per-platform implementations

devicedefender/source/DeviceDefender.cpp
```diff
@@ -114,6 +190,189 @@ namespace Aws
...
+        {
+            CustomMetricNumberFunction func =
+                std::bind(&ReportTask::getCustomMetricMemoryUsage, this, std::placeholders::_1);
+            RegisterCustomMetricNumber("memory_usage", std::move(func));
+        }
+
+        int ReportTask::getCustomMetricMemoryUsage(double *output)
+        {
+// Get the Memory usage from Linux
+#if defined(__linux__) || defined(__unix__)
```
bretambrose, line 339: Similarly, should be moved into aws-c-common with per-platform impl.

bretambrose (commented):
Getting close, just a couple of things (although one will be a pain) stick out.

devicedefender/include/aws/iotdevicedefender/DeviceDefender.h
```diff
@@ -98,6 +203,10 @@ namespace Aws
                 void *cancellationUserdata = nullptr) noexcept;
 
             static void s_onDefenderV1TaskCancelled(void *userData);
+
+            // Holds all of the custom metrics created for this task. These are pointers that will be
+            // automatically created and cleaned by ReportTask when it is destroyed.
+            Crt::Vector<std::shared_ptr<CustomMetricBase>> storedCustomMetrics;
```
bretambrose, line 209: You could better insulate the type definitions by forward declaring CustomMetricBase and then moving its definition plus the definitions of all subclasses to the .cpp file.  The function typedefs would of course stay here.

bretambrose (approved):

PR #407, opened 2022-03-21
Device Advisor CI automation

xiazhvera:
**Description of changes:**
Add the device advisor scripts to enable GitHub Actions to automatically run device advisor test on push

GitHub Setting Changes:
Added Repository secrets: AWS_DATEST_ACCESS_KEY_ID, AWS_DATEST_SECRET_ACCESS_KEY
The secrets are set to aws-sdk-common-runtime user: IotSDKDeviceAdvisorCIAutomation


.github/workflows/ci.yml
```diff
@@ -13,9 +13,12 @@ env:
   PACKAGE_NAME: aws-iot-device-sdk-cpp-v2
   LINUX_BASE_IMAGE: ubuntu-16-x64
   RUN: ${{ github.run_id }}-${{ github.run_number }}
-  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-  AWS_REGION: us-east-1
+#  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
sbSteveK, line 16: should these commented out lines be removed?

sbSteveK (approved):

.builder/actions/build_samples.py
```diff
@@ -5,9 +5,14 @@
 
 class BuildSamples(Builder.Action):
     def run(self, env):
+        skip_da = False
+        skip_samples = False
+        if env.args.cli_config['variables'].get('skip_da', "0") != "0":
```
bretambrose, line 10: If possible, I feel like these should be passed as flags (--skip_da) through the builder and into here.

In fact, why do this at all?  Why not always build the samples?  Running them is separate and based on your variant.

deviceadvisor/script/DATestRun.py
```diff
@@ -0,0 +1,224 @@
+import boto3
```
bretambrose, line 1: Can we look into possibly some shared code refactoring so that the common device/cert/etc... management is a single script somewhere (builder? or dunno) rather than repeating it in each repo?  Language-specific invocation details would still need to be per-repo.

.builder/actions/build_samples.py
xiazhvera: We don't always build the samples because the BYO_CRYPTO mode. With BYO_CRYPTO, the script will not build the tls implementation, and the builder could not build any test sample with it. 

I think you are right, I can change it to use variant.

deviceadvisor/script/DATestRun.py
xiazhvera: That actually would be better. We can add a da test step into builder if possible. I will need to look into the crt-builder for further discussion.

PR #408, opened 2022-03-24
Support mutual TLS using a certificate from a Windows cert store

graebm:
Add the ability to use a client certificate located in a Windows certificate store. Previously, the client certificate and private key had to be passed by filepath or file contents. With this change, certificates and keys stored on TPM devices can be used.

Add new `windows_cert_pub_sub` sample to show this in action.

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -122,7 +122,7 @@ but the private key for mutual TLS is stored on a PKCS#11 compatible smart card 
 
 WARNING: Unix only. Currently, TLS integration with PKCS#11 is only available on Unix devices.
 
-source: `samples/mqtt/pkcs11_pub_sub/main/cpp`
+source: `samples/mqtt/pkcs11_pub_sub/main.cpp`
 
 To run this sample using [SoftHSM2](https://www.opendnssec.org/softhsm/) as the PKCS#11 device:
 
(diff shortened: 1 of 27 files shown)


samples/README.md
```diff
@@ -167,6 +168,68 @@ To run this sample using [SoftHSM2](https://www.opendnssec.org/softhsm/) as the
...
+    ----------                                -------
+    A11F8A9B5DF5B98BA3508FBCA575D09570E0D2C6  CN=AWS IoT Certificate
+    ```
+
+    So this certificate's path would be: "CurrentUser\MY\A11F8A9B5DF5B98BA3508FBCA575D09570E0D2C6"
+
+4) Now you can run the sample:
+
+    ```
+    ./windows-cert-pub-sub.exe --endpoint xxxx-ats.iot.xxxx.amazonaws.com --ca_file AmazonRootCA.pem --cert CurrentUser\My\A11F8A9B5DF5B98BA3508FBCA575D09570E0D2C6
```
sbSteveK, line 231: I always get mixed up but should the slashes be forward slashes in the directory in this command?

sbSteveK (approved):

samples/README.md
graebm: yep, they're supposed to be forward slashes
That's how Windows likes to roll

graebm: It actually DOES need to be backslashes, because that's how Windows likes to roll

PR #419, opened 2022-04-22
Autodocs

graebm:
Automated update the API documentation published to https://aws.github.io/aws-iot-device-sdk-cpp-v2/

**Issue:**
- Docs weren't updated unless someone remembered to do it manually.
- Manually updating docs whenever you update code pollutes the diff, making it hard to review
- Having HTML docs in `main` branch can cause issues.
   - Bloats size.
   - Generated docs might contain licenses we don't want to be redistributing.

**Description of changes:**
- Docs are kept in `docs` branch now
    - New `docs.yml` Github Workflow updates the `docs` branch whenever the `main` branch changes.
    - New `docs-check` job in `ci.yml` checks for warnings/errors in pull-requests.
    - delete the `docs/` folder in `main` and use `.gitignore` to prevent someone from accidentally resurrecting it.


eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -386,7 +386,7 @@ namespace Aws
             virtual bool OnStreamError(Crt::ScopedResource<OperationError> operationError, RpcError rpcError);
         };
 
-        enum AWS_EVENTSTREAMRPC_API ResultType
+        enum ResultType
```
bretambrose, line 389: What is going on in this file?  The import/export macro usage seems really inconsistent.

graebm: I don't know what else is going on in this file, but doxygen was only complaining about this one, and no other enums has an import/export, which makes sense because it's not a symbol a loader would need to find at runtime, it's just a value hardcoded in a header

bretambrose (approved):

PR #430, opened 2022-05-18
Fix segmentation fault in iot client init functions

TwistedTwigleg:
*Description of changes:*

Fixes issue where a segmentation fault will occur if a uninitialized shared pointer is passed in the constructor of the IoT Client classes (Job, Shadow, Secure Tunnel, etc).

___________

### DIFF
--- a/identity/source/IotIdentityClient.cpp
+++ b/identity/source/IotIdentityClient.cpp
@@ -26,7 +26,7 @@ namespace Aws
         {
         }
 
-        IotIdentityClient::operator bool() const noexcept { return *m_connection; }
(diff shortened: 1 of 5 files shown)


TwistedTwigleg:
I also made a slight change to the Jobs sample so there is a 200 millisecond delay before stopping the sample, as otherwise on faster machines it will run too fast and will not output the results of fetching the job details, making it look like it just connected and disconnected. Sleeping for 200 milliseconds fixes the issue and it always prints the job details as expected.

bretambrose (approved):

PR #449, opened 2022-06-17
Secure tunneling stop

bretambrose:
* Adds a shutdown API to allow the device client to work around an async shutdown crash.  The right long-term workaround involves refactoring the C implementation to be safe regardless of API usage pattern.


TwistedTwigleg (approved):
Looks good to me 👍

PR #453, opened 2022-06-28
Add verbosity command to samples

TwistedTwigleg:
*Description of changes:*

Adds the `verbosity` command to all of the samples, making it easier to debug without needing to recompile when you want to change logging verbosity. Also increases consistency with the other SDKs.
(*Also removed `ca_file` from the custom authorizer sample as it is not used in the sample.*)

__________

### DIFF
--- a/samples/utils/CommandLineUtils.cpp
+++ b/samples/utils/CommandLineUtils.cpp
@@ -203,7 +203,7 @@ namespace Utils
         RegisterCommand(
             m_cmd_verbosity,
             "<log level>",
-            "The logging level to use. Choices are 'trace', 'debug', 'info', 'warn', 'error', 'fatal', and 'none'. "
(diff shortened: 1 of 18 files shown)


samples/utils/CommandLineUtils.cpp
```diff
@@ -198,6 +198,52 @@ namespace Utils
             "The password to send when connecting through a custom authorizer (optional)");
     }
 
+    void CommandLineUtils::AddLoggingCommands()
+    {
+        RegisterCommand(
+            m_cmd_verbosity,
+            "<log level>",
+            "The logging level to use. Choices are 'trace', 'debug', 'info', 'warn', 'error', 'fatal', and 'none'. "
```
sbSteveK, line 206: I have to dig around a bit but I feel like these have been "Trace", "Debug", etc w/the first letter capitalized in other SDKs. Might be good to try to be as consistent as possible in the event all the others are this way.

PR #471, opened 2022-08-24
AppVerifier in CI

TwistedTwigleg:
*Description of changes:*

Adds AppVerifier support to CI on Windows so when AppVerifier finds an issue it is surfaced easily. To do this, this PR adds the following:
* New steps in CI to run AppVerifier alongside a new sample, as well as run a Python utility script
* A new PubSub sample called Cycle PubSub
  * This sample will make 3 clients (by default) and will run random operations on them (subscribe, unsubscribe, publish, etc) for a set duration (5 minutes by default).
  * This exercises much more than the other samples in terms of the MQTT client, better allowing AppVerifier to check more of the code being executed.
* A Python utility script with two main functions: Starting the sample and parsing the XML result:
  * Starts the sample given to it. Setups up the environment by downloading the certificate, private key, and endpoint from AWS secrets using `boto3`.
    * Makes launching the sample easier and avoids having to repeat multiple steps over and over again.
  * Will take the XML output from AppVerifier and apply post processing on it
    * Reads the file to determine if AppVerifier found any issues or not. AppVerifier will output XML regardless of whether it found issues, so we cannot rely on the existence of an AppVerifier file alone
    * Takes the non-descriptive error codes and maps them to short descriptions
    * Presents the raw information within the XML file as well if an error occurs

________


bretambrose (approved):
How much work would it be to add app verifier around the unit tests of aws-c-common and aws-c-io?

TwistedTwigleg:
> How much work would it be to add app verifier around the unit tests of aws-c-common and aws-c-io?

Probably not too long, it was mostly just learning how AppVerifier works and getting CI figured out. I would not expect it to take more than a couple days to get both repositories setup, so long as there is a `.exe` to run alongside (if not, then maybe a little more to compile the `.exe` in GitHub actions?)

_______

Thank you for taking a look at the PR! I'll make the necessary changes to fix the merge conflicts 👍

bretambrose:
> > How much work would it be to add app verifier around the unit tests of aws-c-common and aws-c-io?
> 
> Probably not too long, it was mostly just learning how AppVerifier works and getting CI figured out. I would not expect it to take more than a couple days to get both repositories setup, so long as there is a `.exe` to run alongside (if not, then maybe a little more to compile the `.exe` in GitHub actions?)
> 
> Thank you for taking a look at the PR! I'll make the necessary changes to fix the merge conflicts +1

Let's make that a followup task and not hold this back by it, but the aws-c-*-tests.exe (common and io, unsure if there's value elsewhere) should eventually get exercised by that assuming that's reasonable.

TwistedTwigleg:
Merge pending https://github.com/awslabs/aws-crt-cpp/pull/400 being merged and a CRT release made - otherwise CI will fail. Will merge as soon as possible.

TwistedTwigleg:
Merging into `main`...
(*The license scan failure is because we're adding new files so it doesn't have an old license to check against*)

PR #477, opened 2022-09-13
Add generated headers to Greengrass generated code

TwistedTwigleg:
*Description of changes:*

Adds headers to all generated Greengrass code so it is easier to know what is generated from a model, and what is not. Also adds the copyright.

________


sbSteveK (approved):
License check needs fixing. Besides that, looks good!

TwistedTwigleg:
Thanks!

As mentioned in Slack (mentioning it here so it is also attached to PR): The license check is failing because I added new license text, which causes it to go look for the old license but it doesn't exist in the old version, so it errors out.

Merging into `main`...

PR #474, opened 2022-09-06
Add version stamping

TwistedTwigleg:
*Description of changes:*

Adds version stamping to the C++ V2 SDK in the same way it is added to the CRT. Also automatically updates the documentation version string with the latest tagged release.

___________

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #484, opened 2022-10-05
Fix a few things missed in the version stamping PR

TwistedTwigleg:
*Description of changes:*

Fixes a few things missed in the version stamping PR (#474). These fixes include:

* Fixes the `docs.yml` file not reading the version from the `VERSION` file
* Syncs `AwsGetVersion` with the latest changes in the C++ CRT repo
* Removes writing a version file for the root CMake file
* Adds writing a CMake version file for each of the sub-project CMake files.
* Adds stamping all sub-projects with version strings when we make a release
  * Also prints the sub-project version for each sub-project on build

_______


devicedefender/CMakeLists.txt
```diff
@@ -111,6 +114,12 @@ else()
     set(TARGET_DIR "static")
 endif()
 
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(
+        "${CMAKE_CURRENT_BINARY_DIR}/iotdevicedefender-cpp-config-version.cmake"
```
TwistedTwigleg, line 119: The reason I'm using a hard-coded name here, instead of something like `${PROJECT_NAME}` is because the project name is `IotDeviceDefender` while the normal Cmake file is hard-coded to `iotdevicedefender`, so to keep it consistent, this is also hard coded with a similar naming scheme.

devicedefender/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 cmake_minimum_required(VERSION 3.1)
-project(IotDeviceDefender-cpp CXX C)
+
+project(IotDeviceDefender-cpp
+        VERSION ${SIMPLE_VERSION}
```
graebm, line 4: Are these subprojects supposed to support being built standalone? Without being built from the root aws-iot-device-sdk-cpp-v2 root directory? 

If this is built standalone, `${SIMPLE_VERSION}` isn't going to be defined

devicedefender/CMakeLists.txt
```diff
@@ -111,6 +114,12 @@ else()
     set(TARGET_DIR "static")
 endif()
 
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(
+        "${CMAKE_CURRENT_BINARY_DIR}/iotdevicedefender-cpp-config-version.cmake"
+        COMPATIBILITY ExactVersion
```
graebm, line 120: Since the IoT SDK is `1.x` wouldn't we want compatibility to be `SameMajorVersion`?

devicedefender/CMakeLists.txt
TwistedTwigleg: My gut reaction says "no", as they require things like `#include <aws/crt/Api.h>` without setting their CMake file to include `aws-c-common` (for example). However, that said, it looks like *in theory* there is nothing saying it explicitly has to be built using the SDK.

However, it uses things like `aws_use_package` in the CMake file without importing them, in the case of `eventstream_rpc/CMakeLists.txt` - so there is a dependency on being a sub-project to a CMake project that has `aws_use_package` at least.

I think what I will do is try to build a few of the other sub-projects in a standalone build and if they all fail, then I think it is safe to say they need to be built as a CMAKE sub-project to the C++ SDK.

TwistedTwigleg: Update: It appears it is hit and miss. Device Defender can be built standalone and will have issues due to `SIMPLE_VERSION` not existing, while EventStream RPC is dependent on being a sub-project.

I will look into a way to fix this where we can set the version still, but also support standalone builds. 👍

TwistedTwigleg: I fixed it! Now each CMake file has a string that is the last known release version that it uses if `SIMPLE_VERSION` is not set. The version string for each sub-project CMake file is automatically updated as part of running `update-versions.sh`

I checked and it works when the project is standalone (falls back to release string) or when it is a direct sub-project (uses `SIMPLE_VERSION`).

update-version.sh
```diff
@@ -20,6 +20,13 @@ version=$(git describe --tags --abbrev=0)
 version_without_v=$(echo ${version} | cut -f2 -dv)
 echo "${version_without_v}" > VERSION
 
+echo "Updating sub-project versions"
+# First part finds all CMakeList.txt that is 2 spaces deep and no more, limiting it to things like eventstream_rpc/CmakeLists.txt.
+# The second part runs sed (using -exec) and replaces the version string in "set(ENV{_sub_project_version} 1.0.0" with the version of the SDK
+find . -name "CMakeLists.txt" -maxdepth 2 -mindepth 2 -exec sed -i -r -e 's/set(ENV{_sub_project_version} "[^"]*"/set(ENV{_sub_project_version} "'${version_without_v}'"/' {} +
+# This just removes any backup files created by "sed" after calling it.
+find . -name "CMakeLists.txt-r" -maxdepth 2 -mindepth 2 -exec rm {} +
```
graebm, line 28: One of the reasons we went with a `VERSION` file in https://github.com/awslabs/aws-crt-cpp/pull/412 was to avoid the complexity of editing a specific string within a CMakeLists.txt file. It's simpler to read/write a dead-simple file with nothing else in it, than do pattern matching shenanigans like this

Let's do one of these things:
1) Since (none? most?) of the subprojects can't actually be built without building the top-level directory, let's just keep things simple and just assume the top-level project is passing in the version
OR
2) if SIMPLE_VERSION isn't set, read it from "../VERSION" instead of having the value hard-coded here

CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(AwsGetVersion)
 aws_get_version(SIMPLE_VERSION FULL_VERSION GIT_HASH)
 message(STATUS "AWS SDK C++ ${FULL_VERSION}")
+# This is required so we can properly use SIMPLE_VERSION in sub-project CMake files
+set(${SIMPLE_VERSION} ${SIMPLE_VERSION})
```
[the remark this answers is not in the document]
TwistedTwigleg: Strangely, if I didn't do this, then `DEFINED` would not work. After more looking around and experiments, I found the issue and it's sadly just a little typo on my part. I was using `if (DEFINED ${SIMPLE_VERSION})` thinking that is how its supposed to be used, but it's supposed to be used like `if (DEFINED SIMPLE_VERSION)`... 🤦 

This is fixed now.

update-version.sh
TwistedTwigleg: That is fair. I added this because there was discussion on whether these modules can be built outside of the C++ V2 SDK, which I found that some could, though to be honest I doubt anyone does because I cannot see the advantage of doing it that way. Regardless, you can just move some of the sub-projects to a completely different folder and build individually should you so choose. That is why I added the hard-coded string, as a fallback in these cases.

However, I adjusted this using a combination of both things you suggested. Now it checks to see if `SIMPLE_VERSION` exists and if it does, then it uses that, otherwise it checks for the `../VERSION` file to see if it exists and uses it directly if it does. If both of these checks fail, then it just does not version stamp it and prints the version (in the console) out as `unknown`.

This avoids the need to do any pattern matching and still properly version stamps everything in the SDK.

CMakeLists.txt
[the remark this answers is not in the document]
graebm: I truly hate that CMake has its own dumb scripting language.

But CMake started in like 1999 when there weren't a ton of great off-the-shelf options choose from so ... ugh it is what it is

CMakeLists.txt
```diff
@@ -41,18 +41,15 @@ string(REPLACE ";" "${AWS_MODULE_DIR};" AWS_MODULE_PATH "${CMAKE_PREFIX_PATH}${A
...
-)
-
 if (NOT CMAKE_BUILD_TYPE)
     if (NOT WIN32)
         set(CMAKE_BUILD_TYPE "RelWithDebInfo")
     endif()
 endif()
 
+# Tell the CRT the version of the SDK for the metrics string
+set(AWS_IOT_SDK_VERSION ${SIMPLE_VERSION})
```
graebm, line 51: compare this to what is used to be.
shouldn't it be something like `"v${FULL_VERSION}"` ?
so that it's prefixed with "v"
and has a bit of extra junk at the end for dev builds?

devicedefender/CMakeLists.txt
```diff
@@ -1,5 +1,25 @@
...
+
+if (DEFINED _sub_project_version)
+    message("Device Defender version is ${_sub_project_version}")
+    project(IotDeviceDefender-cpp
+            VERSION ${_sub_project_version}
+            LANGUAGES CXX)
+else()
+    message("Device Defender version is unknown")
+    project(IotDeviceDefender-cpp
+        LANGUAGES CXX)
```
graebm, line 21: you can avoid redefining the project in two branches by defining `project()`and then later setting version via 
```
project(IotDeviceDefender-cpp
            LANGUAGES CXX)

if (DEFINED _sub_project_version)
    set_target_properties(IotDeviceDefender-cpp PROPERTIES VERSION ${_sub_project_version})
else()
    message("Device Defender version is unknown")
endif()

Comment (graebm) [reply devicedefender/CMakeLists.txt:21]: But personally, given that they can't be built separately now, I'd kind of prefer we just try to minimize the amount of complex copy pasty code in each one

File: devicedefender/CMakeLists.txt [hunk 8]
```diff
@@ -1,5 +1,25 @@
 cmake_minimum_required(VERSION 3.1)
-project(IotDeviceDefender-cpp CXX C)
+
+if (DEFINED SIMPLE_VERSION)
+    set(_sub_project_version ${SIMPLE_VERSION})
+else()
+    if (EXISTS "../VERSION")
+        file(READ "..//VERSION" version_simple)
+        string(STRIP ${version_simple} version_simple)
+        set(_sub_project_version ${version_simple})
```
graebm, line 9: if you keep this, it could just be 
```
file(READ "..//VERSION" _sub_project_version)
string(STRIP ${version_simple} _sub_project_version)
```

CMakeLists.txt
TwistedTwigleg: Fixed! You were right, the metric string was incorrect for dev builds and needed `v`.

After several iterations, I got it generating the same way every time:
* Desired output (current CRT output): `v1.19.1-11-gfe4ab8e`
* Output without most recent changes: `1.19.1`
* Full version without changes: `v1.19.1-dev+fe4ab8e0`
* Simple version + Git hash (trimmed to 8 characters): `v1.19.1-fe4ab8e0`
  * While this solution provides the correct output, I realized that on a tagged release it will still include the git hash, which we do not want.
* Final solution (Full version with string replace): `v1.19.1-fe4ab8e0`

I also confirmed it gives the proper output if the `.git` folder is missing with the final solution.

graebm (approved):

TwistedTwigleg:
Thanks! Merging into `main`...

PR #525, opened 2023-01-24
GG model update: add client config argument to set async launch mode

xiazhvera:
*Issue #, if available:*

*Description of changes:*
The change creates a client member that setup the async launch mode for the service client. And the client will pass the async config to the service operations.


eventstream_rpc/tests/EchoTestRpcModel.cpp
```diff
@@ -976,9 +976,18 @@ namespace Awstest
         return Aws::Crt::String("awstest#GetAllProducts");
     }
 
-    std::future<GetAllProductsResult> GetAllProductsOperation::GetResult() noexcept
+    std::future<GetAllProductsResult> GetAllProductsOperation::GetResult(bool async_mode) noexcept
```
bretambrose, line 979: I don't think a bool should be used for this.  Bools are not good parameters unless
1. It's absolutely clear what true/false mean without reading the implementation
2. You're 100% sure that there will never be any more possibilities

I don't think this situation passes either of those criteria.

Second, the launch mode is a standard library enum.  Why do we need an if-statement around the entire call?

Also, this isn't backwards compatible; we're adding a required parameter.

Also, it makes sense to me to give the client itself a configuration override so that this doesn't need to be specified over and over.

Finally, from a future expansion point of view, it might be better to have a config object that can be expanded in the future rather than appending more and more parameters with defaults.

bretambrose (changes requested):

TwistedTwigleg (approved):
Looks good to me 👍

PR #528, opened 2023-01-27
CD always get latest version

TwistedTwigleg:
*Description of changes:*

Updates CD to always get the highest git tag version from git tag so even if there are multiple tags on the same day, we always have the newest/latest one.

__________


xiazhvera (approved):

TwistedTwigleg:
Thanks! Merging into `main`...

PR #557, opened 2023-03-31
Secure Tunnel WebSocket Protocol V3 Support

sbSteveK:
Adds support for Secure Tunnel WebSocket Protocol V3


secure_tunneling/source/SecureTunnel.cpp
```diff
@@ -125,12 +168,35 @@ namespace Aws
...
+        //***********************************************************************************************************************
+
+        SendMessageCompleteData::SendMessageCompleteData(
+            enum aws_secure_tunnel_message_type type,
+            Crt::Allocator *allocator) noexcept
+            : m_allocator(allocator)
+        {
+            aws_byte_buf_clean_up(&m_messageTypeStorage);
+            AWS_ZERO_STRUCT(m_messageTypeStorage);
+            struct aws_byte_buf typeBuf = aws_byte_buf_from_c_str(aws_secure_tunnel_message_type_to_c_string(type));
```
TwistedTwigleg, line 190: Does `typeBuf` need to be cleaned after the `aws_byte_buf_init_copy` call? From what I can tell, it looks like it [makes a new copy](https://github.com/awslabs/aws-c-common/blob/194b11eddf2e55854a0994e35115f4ebd988f0b3/source/byte_buf.c#L48) of the data, so I think afterwards you need to erase `typeBuf`. I'm not positive though...

sbSteveK: I think you're right!

secure_tunneling/source/SecureTunnel.cpp
```diff
@@ -728,10 +909,22 @@ namespace Aws
...
+
+            if (secureTunnel->m_OnSendMessageComplete)
+            {
+                std::shared_ptr<SendMessageCompleteData> packet =
+                    std::make_shared<SendMessageCompleteData>(type, secureTunnel->m_allocator);
+                SendMessageCompleteEventData eventData;
+                eventData.sendMessageCompleteData = packet;
+                secureTunnel->m_OnSendMessageComplete(secureTunnel, error_code, eventData);
+            }
+            else if (secureTunnel->m_OnSendDataComplete)
```
xiazhvera, line 927: Debatable: should we use `if` instead of `else if` here? Well, technically if the user already sets OnSendMessageComplete, they would not and should not use OnSendDataComplete in this case. I feel it is better to make sure the callbacks work independently?  Or at least mentioned in the comments that if you've set OnSendMessageComplete, OnSendDataComplete would no longer be trigged.

xiazhvera (approved):
Looks good to me, trivial comments.

TwistedTwigleg (approved):

Issue #551, opened 2023-03-23
Issue with sample project build command due to -B option in cmake version


Kta-M:
### Describe the issue

In the README.md file, it is mentioned that cmake version 3.1+ is required for building the project. While building the main project, there were no issues. However, I encountered a problem while trying to build the sample project.

The issue seems to be with the -B option in the cmake command used for building the sample project. This option was introduced in cmake version 3.13, which is not compatible with the stated version requirement of cmake 3.1+.

### Links

https://cmake.org/cmake/help/latest/manual/cmake.1.html#generate-a-project-buildsystem

TwistedTwigleg:
Thanks for reporting this issue. The sample README now includes instructions on how to build the samples without the use of the `-B` command, which should work for CMake 3.1+.

Thanks again!

PR #563, opened 2023-04-12
Sample and command line refactor

TwistedTwigleg:
*Description of changes:*

Refactors samples not to rely on `cmdUtils` to setup MQTT clients, to connect/disconnect for connect samples, and overall just makes `cmdUtils` go back to command line parsing and nothing more.

Also contains the following minor changes:
* Renames `Identity` folder to `fleet_provisioning` for better visibility.
* Adjusts all samples to register all command line commands and get the data from said commands in a single function, rather than doing it individually in each sample.
  * This also fixed a few minor inconsistencies that slipped through: like `region` being used in a couple samples instead of `signing_region`.
* Removed RawConnect sample entirely
  * The AWS IoT MQTT Builder should be used instead. If a sample on using the CRT is wanted/needed, it should be in the CRT repository.
* Fixes possible Cross-SDK repository communication for PubSub by appending UUIDs to topics
* Adjusts comments and formatting for a more consistent layout and easier reading.

Edit: Also adjusted sample README to mention how to build in CMake without `-B`, which should fix the following issue: https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/551

__________


sbSteveK (approved):

xiazhvera:
Out of curiosity: Is there any reason why we use // instead of /**/ for most of the comments? I figured that you did that change on purpose. Just curious why we did that? Otherwise I feel this change makes the comments less noticeable. Also useing /**/ for block comments seems more common in the library.

TwistedTwigleg:
I changed the comments from `/* */` to simply `//` for two reasons:
* First reason is that `//` is supported in C++, while in C it's *technically* not supported and is part of the compiler. Using comments like `/* */` are the C-style of adding comments.
* Second reason is that it reduces the lines needed for comments by at least 2 for the majority of comments. Since I was already exploding the lines of code for each sample, I thought that reducing the lines by minimizing the comments might be a helpful compromise to help keep the samples a bit more lean.

However, I do not feel strongly about it either way and can totally change `//` back to `/* */` if that would be preferred. I do agree that it makes the comments a bit less visible, but I thought the potential tradeoff was worth it, in terms of reducing some of the lines of code while using C++ style comments.

PR #572, opened 2023-04-19
optional tls settings on secure tunnel creation

sbSteveK:
* Add option to provide tls options when building a new secure tunnel

### DIFF
--- a/secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
+++ b/secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
@@ -617,6 +617,16 @@ namespace Aws
                 const std::string &endpointHost); // Make a copy and save in this object
 
             /* Optional members */
+
+            /**
+             * Sets TLS options to be used by secure tunnel connection.
+             *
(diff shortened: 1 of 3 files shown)


secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
```diff
@@ -1130,6 +1146,7 @@ namespace Aws
                 aws_secure_tunneling_local_proxy_mode localProxyMode,
                 const std::string &endpointHost,
 
+                Crt::Io::TlsConnectionOptions *tslOptions,
```
xiazhvera, line 1149: Trivial: `tlsoOptions`

xiazhvera (approved):
The change itself looks good to me, while I feel we should document that if the tlsOptions is set manually, then the `root_ca` configuration will be overwritten.

secure_tunneling/include/aws/iotsecuretunneling/SecureTunnel.h
sbSteveK: Good call. Will add it to the documentation

PR #616, opened 2023-08-16
Generated jobs client with enum Fix

sbSteveK:
Fixed headers for jobs using enums to correctly build with shared libs on Windows.


bretambrose (changes requested):
This does not look right.  Enums should not shift position.  And there should be no other code changes than the conversion function declspec macro position

PR #618, opened 2023-08-17
Fix enums in jobs

sbSteveK:
Fix enums in jobs


bretambrose (commented):
Can we also add a CI job that does a shared object Windows build?

PR #644, opened 2023-10-09
Readme cleanup

jmklix:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,6 @@
 # AWS IoT Device SDK for C++ v2
 
-This document provides information about the AWS IoT device SDK for C++ V2.
-
-If you have any issues or feature requests, please file an issue or pull request.
-
-This SDK is built on the AWS Common Runtime, a collection of libraries
-([aws-c-common](https://github.com/awslabs/aws-c-common),
-[aws-c-io](https://github.com/awslabs/aws-c-io),
-[aws-c-mqtt](https://github.com/awslabs/aws-c-mqtt),
-[aws-c-http](https://github.com/awslabs/aws-c-http),
-[aws-c-cal](https://github.com/awslabs/aws-c-cal),
-[aws-c-auth](https://github.com/awslabs/aws-c-auth),
-[s2n](https://github.com/awslabs/s2n)...) written in C to be
-cross-platform, high-performance, secure, and reliable. The libraries are bound
-to C++ by the [aws-crt-cpp](https://github.com/awslabs/aws-crt-cpp) package.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,29 +1,15 @@
...
 __Jump To:__
 
 * [Installation](#installation)
 * [Samples](./samples)
 * [Getting Help](#getting-help)
 * [FAQ](./documents/FAQ.md)
-* [Giving Feedback and Contributions](#giving-feedback-and-contributions)
+* [API Docs](https://aws.github.io/aws-iot-device-sdk-cpp-v2/)
 * [MQTT5 User Guide](./documents/MQTT5_Userguide.md)
-* [Secure Tunnel User Guide](./documents/Secure_Tunnel_Userguide.md)
```
xiazhvera, line 26: Would we like to keep the link to Secure Tunnel User guide?

sbSteveK: +1. Alternately, we could move the secure tunnel guide elsewhere but we should have a link somewhere to it.

README.md
```diff
@@ -54,6 +38,11 @@ cd ..
 # If working with multiple SDKs, using a SDK-specific name is helpful.
```
sbSteveK, line 38: Not part of this PR but this line feels a little unclear and maybe unnecessary.

README.md
```diff
@@ -64,25 +53,11 @@ cmake --build . --target install
 ```
 
 #### Windows
-
 ``` sh
-# Create a workspace directory to hold all the SDK files
```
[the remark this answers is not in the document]
sbSteveK: we probably don't need to go so far as instructing someone to create a workspace directory though. I think the use of the word "workspace" could cause confusion between this directory and the "absolute path to sdk-workspace" since I don't believe they're both pointing to the same directory.

README.md
sbSteveK: I think "git clone --recursive" makes is unnecessary to go into the folder and run "git submodule update --init --recursive" so we can probably cut out lines 49 - 52. Unless there's some reason why the submodules wouldn't be properly updated with the initial "git clone --recursive" command.

README.md
```diff
@@ -32,10 +32,8 @@ cd sdk-workspace
 git clone --recursive https://github.com/aws/aws-iot-device-sdk-cpp-v2.git
 # Ensure all submodules are properly updated
 cd aws-iot-device-sdk-cpp-v2
```
sbSteveK, line 34: no need to change into this directory and change back out of this directory.

sbSteveK (approved):
cd into dir and cd .. ship

xiazhvera (approved):
The Codebuild was failed because of ubuntu-toolchain. Rerun the CI should fix it.

PR #660, opened 2023-12-01
Shadow Update tests for CI

alfred2g:
Shadow update tests


servicetests/tests/ShadowUpdate/main.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+#include "../../../samples/utils/CommandLineUtils.h"
+
+using namespace Aws::Crt;
+using namespace Aws::Iotshadow;
+
+void changeShadowValue(Aws::Crt::String thingName, String property, String value);
+void changeNamedShadowValue(String thingName, String property, String value, String shadowName);
+
+IotShadowClient *shadowClient = nullptr;
```
bretambrose, line 42: Seems messy to have a random global variable here.  Why not a local unique/shared_ptr in main?

servicetests/tests/ShadowUpdate/main.cpp
```diff
@@ -0,0 +1,331 @@
...
+int main(int argc, char *argv[])
+{
+    printf("Starting shadow update test\n");
```
bretambrose, line 45: There's a lot of debug-esque output throughout here.  It should either be removed (if it's not necessary) or made into logging statements so that CRT logs from runs give a full picture of what's going on.

servicetests/tests/ShadowUpdate/main.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+    /**
+     * In a real world application you probably don't want to enforce synchronous behavior
+     * but this is a sample console application, so we'll just do that with a condition variable.
+     */
+    std::promise<bool> connectionCompletedPromise;
+    std::promise<void> connectionClosedPromise;
+    std::shared_ptr<Aws::Crt::Mqtt::MqttConnection> connection;
+
+    Aws::Iot::Mqtt5ClientBuilder *builder = Aws::Iot::Mqtt5ClientBuilder::NewMqtt5ClientBuilderWithMtlsFromPath(
```
bretambrose, line 64: This test seemingly supports both 3 and 5 and yet we have a 5 builder unconditionally created and just sitting at top-level scope.

Can we partition the setup out into a create_client helper which in turn conditionally invokes a create_3_based_client and create_5_based_client helper as appropriate?

servicetests/tests/ShadowUpdate/main.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+    // Check if the builder setup correctly.
+    if (builder == nullptr)
+    {
+        printf(
+            "Failed to setup mqtt5 client builder with error code %d: %s", LastError(), ErrorDebugString(LastError()));
+        return -1;
+    }
+    // Create Mqtt5Client
+    std::shared_ptr<Aws::Crt::Mqtt5::Mqtt5Client> client5;
```
bretambrose, line 75: Same here, if 3 and 5 are both supported then why put a 5 client out here?

bretambrose (changes requested):

servicetests/tests/ShadowUpdate/main.cpp
alfred2g: I need that client to call stop connection for mqtt5
and need the connection pointer to call close connection for mqtt3

servicetests/tests/ShadowUpdate/main.cpp
```diff
@@ -0,0 +1,352 @@
...
+    state.Reported = reported;
+
+    UpdateShadowRequest request;
+    request.ThingName = thingName;
+    request.State = state;
+    Aws::Crt::UUID uuid;
+    request.ClientToken = uuid.ToString();
+
+    std::promise<void> shadowCompletedPromise;
+    auto publishCompleted = [thingName, value, &shadowCompletedPromise](int ioErr) {
```
bretambrose, line 290: We're still not verifying anything actually works beyond connecting.

alfred2g: true, because we are verifying in the launch script, that the job did execute

bretambrose (approved):
Nothing significant, but there's printfs, fprintf to stdout, fprintf to stderr, all sprinkled around.  Can we just use one?

PR #696, opened 2024-03-20
Doc supported arch

alfred2g:
Add supported architectures to the Readme


README.md
```diff
@@ -15,6 +15,35 @@ __Jump To:__
...
+- x86\_64
+- x86
+#### Windows
+- Win32
+- x64
+- ARM64
+#### Mac
+- Apple Silicone (M1 and higher)
+- Apple Intel Chips (x86\_64)
+#### IOS
```
xiazhvera, line 42: I'm not confident about the iOS and Android. I know we have cross compile android and iOS in crt-cpp, but as we didnt test it in SDKs, we probably should not mention it here. Unless we have confirmed they worked with the service clients?

README.md
```diff
@@ -15,6 +15,35 @@ __Jump To:__
...
+#### Linux
+- manylinux2014-x64
+- manylinux2014-x86
+- al2-x64
+- alpine-3.16-x64
+- alpine-3.16-x86
+- alpine-3.16-armv6
+- alpine-3.16-armv7
+- alpine-3.16-arm64
+- raspberry pi
```
xiazhvera, line 29: We probably would like to mention raspberry pi **bullseye** as that is the only raspberry pi os we tested against.

README.md
```diff
@@ -15,6 +15,35 @@ __Jump To:__
 
 ## Installation
 
+### Supported Architectures
```
tkc4106, line 18: Please move the "**Supported Architectures**" section outside of "**Installation**" section. It should be on it own. Keep it at the top.

xiazhvera (approved):

PR #694, opened 2024-03-15
Clean and document eventstream

sfod:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #700, opened 2024-03-27
Add comments and readme file for eventstream module

sfod:


eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -608,7 +788,7 @@ namespace Aws
             }
 
             /**
-             * @return true if the instance is in a valid state, false otherwise.
+             * @return true if the connection is open, false otherwise.
```
bretambrose, line 791: The original comment captures the intent of the API.  The new comment says what it actually does internally.  Not sure if this is an improvement or makes things less clear.

eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -504,18 +631,48 @@ namespace Aws
...
+             */
             std::future<RpcError> Activate(
                 const AbstractShapeBase *shape,
                 OnMessageFlushCallback onMessageFlushCallback) noexcept;
             std::future<RpcError> SendStreamEvent(
                 AbstractShapeBase *shape,
                 OnMessageFlushCallback onMessageFlushCallback) noexcept;
+
+            /**
+             * Returns the canonical model name associated with this context across any client language.
```
bretambrose, line 670: maybe replace context with operation

eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
```diff
@@ -299,23 +307,63 @@ namespace Aws
...
                 MessageType messageType,
                 uint32_t messageFlags,
                 OnMessageFlushCallback onMessageFlushCallback) noexcept;
+
+            /**
+             * Check if the continuation has been closed.
+             * @return True if the continuation has been closed, false otherwise.
+             */
             bool IsClosed() noexcept;
             void Release() noexcept;
```
bretambrose, line 355: Release probably needs a comment

sfod: This method doesn't have implementation. I removed this method in my original attempt to document this module: https://github.com/aws/aws-iot-device-sdk-cpp-v2/pull/694/files#diff-d1487c1bd42c87e8c6d157086720cf4623bfaa5146fc1060072df42b1471cffcL318.
Eventually I decided to split the original PR into multiple: one for docs, one for cleaning up the code, one for fixing various issues. Because it started turn into unreviewable mess pretty quick.

Saying that, I don't have a strong opinion on the matter, so I can remove unimplemented methods (there are more than one) if you think it's the right thing to do here.

eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
sfod: This operator returns false for a closed connection or for a connection awaiting acknowledge, which are valid states. To me, the original comment feels misleading.

eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
bretambrose: That's fine then.

eventstream_rpc/include/aws/eventstreamrpc/EventStreamClient.h
bretambrose: Hmm, it's unfortunate that this is behaviorally different than every other bool cast operator in the CRT, but I supposed we can't change it now.

sfod: The comment can be rephrased to something like this: `return true if the instance is valid and ready to send data, false otherwise`.

bretambrose (approved):

PR #710, opened 2024-04-12
Update codegen

sfod:


sbSteveK (approved):
Trivial: every get and set combo seems to have the exact same documentation without differentiating that one is setting and one is getting the value. It's pretty obvious what a `get_thing` vs. `set_thing` is doing but just weird to see identical definitions.

sfod:
The doc messages for setters and getters come from the corresponding field descriptions. I tried to prepend these messages with `Set ` and `Get `, but it made some of them to look worse. So I decided to add these docs as-is.

PR #714, opened 2024-04-16
Add USE_EXTERNAL_DEPS_SOURCES cmake option

sfod:
Add a new cmake option, `USE_EXTERNAL_DEPS_SOURCES`. When building dependencies is turned off, this new option will signal that the dependencies are provided not as installed libraries, but rather as subprojects (i.e. added previously using `add_subdirectory` command).


xiazhvera (approved):

CMakeLists.txt
```diff
@@ -70,8 +75,27 @@ if (BUILD_DEPS)
     endif ()
     set(BUILD_TESTING ${BUILD_TESTING_PREV})
 else()
+    if (USE_EXTERNAL_DEPS_SOURCES)
+        function(validate_deps_sources dep_name)
+            if (NOT ${dep_name}_SOURCE_DIR)
```
archigup, line 80: This variable only exists if FetchContent is used
Probably should not block on it existing, and just use the targets. The user will get a error saying the target is not defined in that case.

sfod: This variable is set by `add_subdirectory` (which is used by FetchContent): https://cmake.org/cmake/help/v3.14/variable/PROJECT-NAME_SOURCE_DIR.html

If we remove this check, then some error will appear down the code, most probably about a module or a library not found. I think it's more user friendly to notify user what they supposed to do.

PR #721, opened 2024-05-28
Fix macos CI job

sfod:
- Add macos-13 image to OSX job.
- Use Python virtual env to install packages. 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -352,6 +352,7 @@ jobs:
           aws-region: ${{ env.AWS_DEFAULT_REGION }}
       - name: run MQTT3 PubSub sample
         run: |
+          source .venv/bin/activate
           python3 ${{ env.CI_UTILS_FOLDER }}/run_sample_ci.py --file ${{ env.CI_SAMPLES_CFG_FOLDER }}/ci_run_pubsub_cfg.json
 #      - name: run PKCS12 sample
 #        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -325,7 +325,7 @@ jobs:
           python builder.pyz build -p ${{ env.PACKAGE_NAME }} --cmake-extra=-DBUILD_SHARED_LIBS=ON
 
   osx:
-    runs-on: macos-latest
+    runs-on: macos-13
```
xiazhvera, line 328: As Github has more OS version support, we can run on multiple OS versions using the matrix, like what we did for linux. 

```suggestion
    runs-on: ${{ matrix.runner }}
    strategy:
      matrix:
        runner:
          - macos-13
          - macos-latest
```

sfod: `macos-latest` currently fails. In the future, yes, that's a good idea.

xiazhvera: I saw a workaround there https://github.com/orgs/Homebrew/discussions/5159
It is kind of hack, while I think it should be fine as we just use it in CI tests.

.github/workflows/ci.yml
```diff
@@ -337,7 +342,7 @@ jobs:
           ./builder build -p ${{ env.PACKAGE_NAME }} --spec=downstream
       - name: Running samples in CI setup
         run: |
-          python3 -m pip install boto3
+          python3 -m pip install boto3 --break-system-packages
```
sbSteveK, line 345: Do we know why "--break-system-packages" is required for latest MacOS version? Would it be better to instead create and use a venv? Slightly worried that forcing the installation could cause a harder to track down problem in the future depending on what boto3 is overriding and/or the conflicts this is ignoring.

sbSteveK (approved):
Remove macos-latest and  --break-system-packaegs and ship. Optionally add macos-14.

PR #733, opened 2024-07-11
Fix race conditions in fleet provisioning samples

sfod:

### DIFF
--- a/samples/fleet_provisioning/fleet_provisioning/main.cpp
+++ b/samples/fleet_provisioning/fleet_provisioning/main.cpp
@@ -59,8 +59,7 @@ struct CreateCertificateContext
     std::promise<void> pubAckPromise;
     std::promise<void> acceptedSubAckPromise;
     std::promise<void> rejectedSubAckPromise;
-    std::promise<void> tokenReceivedPromise;
-    String token;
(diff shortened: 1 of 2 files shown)


samples/fleet_provisioning/fleet_provisioning/main.cpp
```diff
@@ -21,87 +27,82 @@
...
-    String csrFile;
+/**
+ * Auxiliary structure for holding data used when creating a certificate.
+ */
+struct CreateCertificateContext
+{
+    std::promise<void> pubAckPromise;
+    std::promise<void> acceptedSubAckPromise;
+    std::promise<void> rejectedSubAckPromise;
+    std::promise<void> tokenReceivedPromise;
```
bretambrose, line 62: Technically could probably merge token into the tokenReceivedPromise by changing it to std::promise<String> but not a big deal.

bretambrose (approved):

samples/fleet_provisioning/fleet_provisioning/main.cpp
sfod: That's a good improvement, I'll do it.

PR #763, opened 2024-11-05
Request response samples

bretambrose:
* Move old shadow, jobs, and identity samples to deprecated folder
* Add new samples based on the v2 clients


samples/fleet_provisioning/provision-basic/main.cpp
```diff
@@ -0,0 +1,191 @@
...
+    connectedWaiter.get_future().wait();
+
+    /*
+     * Step 1: Invoke the CreateKeysAndCertificate API to get a key pair signed by AmazonRootCA1
+     */
+    CreateKeysAndCertificateRequest createKeysRequest;
+
+    // A simple `std::promise<Result>` can be used here. But to support old VS compilers requiring a template parameter
+    // be default-constructible, we wrap Result into Optional.
+    std::promise<Aws::Crt::Optional<CreateKeysAndCertificateResult>> createKeysResultPromise;
```
sfod, line 126: We hit here an old VS bug: https://developercommunity.visualstudio.com/t/c-shared-state-futuresstate-default-constructs-the/60897
Alternatively, we could make [`Result`](https://github.com/awslabs/aws-crt-cpp/blob/v0.32.4/include/aws/iot/MqttRequestResponseClient.h#L257) to be default-constructible, but a default value doesn't make much sense for it.

PR #780, opened 2025-02-21
Fix shadow samples in CI

sfod:
*Issue #, if available:*

When multiple instances of shadow samples are running in CI, they use the same thing and interfere with each other causing crashes:
```
Received shadow document.
Shadow contains "color". Updating local value to "Shadow_Value_0"...

stderr:
terminate called after throwing an instance of 'std::future_error'
  what():  std::future_error: Promise already satisfied
```

*Description of changes:*

Handle multiple incoming shadow messages.


bretambrose (approved):
Can't wait to delete these.

PR #797, opened 2025-04-24
CI small fix

sbSteveK:
Deleting an IoT thing before its principals are all released was causing CI to fail. Wait for the principals to be released before moving onto deleting.

### DIFF
--- a/utils/ci_iot_thing.py
+++ b/utils/ci_iot_thing.py
@@ -3,8 +3,46 @@
 
 import sys
 
-import boto3
+import boto3, time
+from botocore.exceptions import ClientError, WaiterError
 
+class ThingDetachedWaiter:
+    """
+    Wait until principal (cert or Cognito identity) is detached from a thing.
+    Raise WaiterError after timeout seconds.
+    """
+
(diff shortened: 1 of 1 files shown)


utils/ci_iot_thing.py
```diff
@@ -65,12 +105,24 @@ def delete_iot_thing(thing_name, region):
             certificate_id = principal.split("/")[1]
             iot_client.detach_thing_principal(thingName=thing_name, principal=principal)
             iot_client.update_certificate(certificateId=certificate_id, newStatus='INACTIVE')
-            iot_client.delete_certificate(certificateId=certificate_id, forceDelete=True)
+            cert_ids.append(certificate_id)
     except Exception:
-        print(f"ERROR: Could not delete certificate for IoT thing {thing_name}, probably thing does not exist",
+        print("ERROR: Could not detatch principal or set its certificate to INACTIVE for {thing_name}, probably thing does not exist",
```
sfod, line 110: trivial: f-string is needed here for interpolation. I see that this change came from java-v2 version of this script that doesn't have the `f` literal. Not a big deal, but can be fixed here, I think

sfod (approved):

PR #816, opened 2025-06-19
Eventstream Refactor

bretambrose:
This PR is a substantial rewrite of the eventstream RPC bindings as well as the code-generated clients for eventstream-based services.  This refactor was necessitated by a variety of deadlock and race condition problems with the original implementation.  The complexity of the original implementation made targeted fixes nearly impossible to apply.

## Refactor Goals
* No blocking in destructors.  In order to try and maintain behavioral compatibility with the previous implementation, we try and synchronously simulate the asynchronous events that would happen during a blocking destroy.
* Simplified synchronization model where
* * User callbacks are never invoked inside a lock
* * C APIs are never invoked inside a lock

## Public API Changes

The original implementation exposed a large amount of unnecessary details in the public API.  As part of the refactor, we make a number of publicly visible changes that, while technically breaking, we believe should not be user-impacting.  We consider a change to be user-impacting if it is a breaking change to a type that is used during service client interaction.

We detail each change below as well as the reasoning why we think making this change is safe.  Obviously, if you were mocking out any of these changed type contracts, then they will be breaking.

* All `OperationModelContext` subclasses have been made private.  These types were used internally by the service model and there is no reason to expose them.
* `ContinuationCallbackData` removed.  Was not user-facing.  Unneeded in refactor
* `ClientContinuationHandler` - Public functions that were only for internal use have been removed.  Class now useless but has been retained in case users were tracking operations by it.
* `ClientContinuation` - Internal type that has been re-implemented as the private type `ClientContinuationImpl`
* `ClientOperation`
* * Constructor type signature has changed - This type is only constructed internally by generated code
* * `GetOperationResult` API removed - This function could not be called externally without triggering exceptions by multi-consuming a promise's future
* * `WithLaunchMode` - This function persists but no longer does anything useful.  Launch mode is no longer relevant to the processing of operations and was a mistake to include originally.
* `ClientConnection` - This is an internal type used by generated service clients.
* * Constructor signature changed.
* * `SendPing` and `SendPingResponse` removed.  
* * `Connect` and `NewStream` signatures changed.  
* * bool operator removed

## Additional Changes

We now launch an EchoTest RPC server in CI and run a much larger suite of tests against it.


eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -674,500 +503,799 @@ namespace Aws
...
+            s_fillNativeHeadersArray(messageAmendment.GetHeaders(), headersArray, m_allocator);
+
+            msg_args->headers = (struct aws_event_stream_header_value_pair *)headersArray->data;
+            msg_args->headers_count = messageAmendment.GetHeaders().size();
+            msg_args->payload = messageAmendment.GetPayload().has_value()
+                                    ? (aws_byte_buf *)(&(messageAmendment.GetPayload().value()))
+                                    : nullptr;
+            msg_args->message_type = AWS_EVENT_STREAM_RPC_MESSAGE_TYPE_CONNECT;
+            msg_args->message_flags = 0U;
         }
```
sfod, line 1036: `messageAmendment.m_payload` will be destroyed on the block end, so `msg_args->payload` will cause use-after-free

eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -1260,377 +1388,1078 @@ namespace Aws
...
+            // The continuation is closed
+            Closed,
+        };
+
+        struct ContinuationSharedState
+        {
+            ContinuationSharedState();
+
+            ContinuationStateType m_currentState;
+            ContinuationStateType m_desiredState;
```
sfod, line 1487: `m_desiredState` seems never used

eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -1260,377 +1388,1078 @@ namespace Aws
...
-            (void)rpcError;
-            /* Note: Always returning true implies that the stream should close
-             * as a result of encountering this error. */
-            return true;
+            aws_client_bootstrap_release(m_clientBootstrap);
         }
 
-        void StreamResponseHandler::OnStreamEvent(Crt::ScopedResource<AbstractShapeBase> response) {}
+        // We use a task to release the C continuation to make it impossible to trigger the termination callback
+        // in a callstack that includes ClientContinationImpl methods or state.
```
sfod, line 1645: Typo in ClientContinationImpl

eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -1260,377 +1388,1078 @@ namespace Aws
...
+                                    static_cast<OperationError *>(shape.release()),
+                                    [allocator](OperationError *shape) { Crt::Delete(shape, allocator); });
+
+                                shouldClose = streamHandler->OnStreamError(
+                                    std::move(errorResponse), {EVENT_STREAM_RPC_SUCCESS, 0});
+                                break;
+                            }
+
+                            default:
+                                AWS_FATAL_ASSERT(false);
```
sfod, line 2332: This assert can be fired when `ClientContinuationImpl::Close` is called in parallel.
This happens when `Close` sets `m_sharedState.m_currentState = ContinuationStateType::PendingClose`, which leads to `isResponse == false`.
I think checking for `PendingClose` when determining the value for `isResponse` should resolve this issue.

eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -1260,377 +1388,1078 @@ namespace Aws
...
+        MessageResult ClientContinuationImpl::DeserializeRawMessage(
+            const struct aws_event_stream_rpc_message_args *rawMessage,
+            bool shouldBeActivationResponse) noexcept
+        {
+            MessageResult result;
+
+            Crt::List<EventStreamHeader> continuationMessageHeaders;
+            for (size_t i = 0; i < rawMessage->headers_count; ++i)
+            {
+                continuationMessageHeaders.emplace_back(EventStreamHeader(rawMessage->headers[i], m_allocator));
```
sfod, line 2123: `EventStreamHeader` explicitly specified here causes unnecessary move constructor call (which actually copies data). Not big deal, but if it can be easily avoided, why not:
```c++
continuationMessageHeaders.emplace_back(rawMessage->headers[i], m_allocator);
```

There are few other places where type shouldn't be used with `emplace_back`.

You can take a look at [godbolt example](https://godbolt.org/z/a31K47a78)

eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -674,500 +503,799 @@ namespace Aws
...
+            AWS_LOGF_INFO(
+                AWS_LS_COMMON_GENERAL, "EventStreamClient2 - Host name %d bytes long", (int)hostName.length());
+            AWS_LOGF_INFO(AWS_LS_COMMON_GENERAL, "EventStreamClient2 - Host name : %s", hostName.c_str());
+
+            return localFuture;
+        }
+
+        bool ClientConnectionImpl::IsOpen() const noexcept
+        {
+            std::lock_guard<std::mutex> lock(const_cast<ClientConnectionImpl *>(this)->m_sharedStateLock);
```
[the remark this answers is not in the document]
bretambrose: How?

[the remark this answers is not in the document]
sfod: Declare the `m_sharedStateLock` member like this:
```
mutable std::mutex m_sharedStateLock;
```

Then we'll be able to change `m_sharedStateLock` in const functions:
```
        bool ClientConnectionImpl::IsOpen() const noexcept
        {
            std::lock_guard<std::mutex> lock(m_sharedStateLock);
            return m_sharedState.m_currentState == ClientState::Connected;
        }
```

eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -674,500 +503,799 @@ namespace Aws
...
             {
-                aws_array_list_clean_up(&headersArray);
+                // Rollback our optimistic state update
+                MoveToDisconnected({EVENT_STREAM_RPC_CONNECTION_SETUP_FAILED, aws_last_error()});
             }
 
-            if (errorCode)
+            AWS_LOGF_INFO(
+                AWS_LS_COMMON_GENERAL, "EventStreamClient2 - Host name %d bytes long", (int)hostName.length());
+            AWS_LOGF_INFO(AWS_LS_COMMON_GENERAL, "EventStreamClient2 - Host name : %s", hostName.c_str());
```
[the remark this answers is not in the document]
bretambrose: Right was leftover from debugging

eventstream_rpc/source/EventStreamClient.cpp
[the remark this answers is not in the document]
bretambrose: TIL

eventstream_rpc/source/EventStreamClient.cpp
bretambrose: Good thing we never use that I guess

eventstream_rpc/source/EventStreamClient.cpp
bretambrose: Not following.  If `isResponse` is false then the deserialize will yield a `Stream` value.

sfod: You're right, I missed this

eventstream_rpc/tests/EventStreamClientTest.cpp
```diff
@@ -86,457 +95,2148 @@ class TestLifecycleHandler : public ConnectionLifecycleHandler
...
+            auto causeServiceError = client.NewCauseServiceError();
+            CauseServiceErrorRequest causeServiceErrorRequest;
+
+            auto requestFuture = causeServiceError->Activate(causeServiceErrorRequest, s_onMessageFlush);
+            requestFuture.wait();
+            auto result = causeServiceError->GetResult().get();
+            ASSERT_FALSE(result);
+            ASSERT_NOT_NULL(result.GetOperationError());
+            auto error = result.GetOperationError();
+            const auto &errorMessage = error->GetMessage().value();
```
sfod, line 908: `errorMessage` will be a dangling ref here, because `error->GetMessage()` returns a temporary object, so then `T &&value() &&` is called, i.e. we deal with [xvalue](https://en.cppreference.com/w/cpp/language/value_category.html#xvalue). And xvalue's lifetime can't be extended.

This should resolve the issue:
```c++
const auto &errorMessageOpt = error->GetMessage();
const auto &errorMessage = errorMessageOpt.value;
```

eventstream_rpc/source/EventStreamClient.cpp
```diff
@@ -674,963 +498,1905 @@ namespace Aws
...
+             */
+            bool m_continuationValid;
+        };
+
+        ClientContinuationImpl::ClientContinuationImpl(
+            Aws::Crt::Allocator *allocator,
+            struct aws_client_bootstrap *bootstrap,
+            struct aws_event_stream_rpc_client_connection *connection) noexcept
+            : m_allocator(allocator), m_clientBootstrap(aws_client_bootstrap_acquire(bootstrap)),
+              m_connectionEventLoop(aws_event_stream_rpc_client_connection_get_event_loop(connection)), m_sharedState(),
```
sfod, line 1523: I took a look at the CI sanitizer errors, and it's shutdown issue: when `ClientOperation` sends its close message on destruction, `channel` is already destroyed.
I didn't dig deep into this issue. But there is a chance that acquiring the event loop could be the right direction.

sfod (approved):
Beside a couple trivial things, looks good

PR #829, opened 2025-08-28
Sample CI Cleanup

sbSteveK:
- Disentangle Samples from CI
- Remove deprecated samples and update documentation
- Remove Device Advisor Testing

### DIFF
--- a/.builder/actions/build_samples.py
+++ b/.builder/actions/build_samples.py
@@ -26,7 +26,6 @@ class BuildSamples(Builder.Action):
             'samples/fleet_provisioning/provision-csr',
             'samples/jobs/jobs-sandbox',
             'samples/mqtt5/mqtt5_pubsub',
-            'samples/mqtt5/mqtt5_shared_subscription',
             'samples/secure_tunneling/secure_tunnel',
             'samples/secure_tunneling/tunnel_notification',
             'samples/shadow/shadow-sandbox',


samples/CMakeLists.txt
```diff
@@ -9,27 +9,7 @@ add_subdirectory(fleet_provisioning/provision-csr)
...
-add_subdirectory(mqtt/basic_connect)
-add_subdirectory(mqtt/custom_authorizer_connect)
-add_subdirectory(mqtt/pkcs11_connect)
-add_subdirectory(mqtt/pkcs12_connect)
-add_subdirectory(mqtt/websocket_connect)
-add_subdirectory(mqtt/windows_cert_connect)
-add_subdirectory(mqtt/x509_credentials_provider_connect)
-add_subdirectory(mqtt/cognito_connect)
-add_subdirectory(mqtt5/mqtt5_pubsub)
-add_subdirectory(mqtt5/mqtt5_shared_subscription)
```
sfod, line 21: Removed by accident?

sbSteveK: Intentional. We had a brief discussion and the sample is using 3 clients to subscribe and publish to a shared subscription topic. Coverage in the docs is enough for shared subscriptions and the sample itself was overly confusing. It could be argued that we rewrite it but it doesn't feel worth the effort.

sbSteveK: Direct discussion offline and yes, shared subscription can be removed but mqtt5_pubsub should not be. Fixing!

